Issue #111, opened 2026-04-15
Implement the Order Management System (OMS) with bracket, OCO, and TWAP/VWAP/POV algos


SevFle:
https://github.com/SevFle/WedPilot/issues/111

## Feature Task

Build professional planner integration (EP-009):

### Requirements
- Planner profile system with certifications and ratings
- Planner search and discovery
- Coaching session booking (calendar integration)
- Chat interface for planner communication
- Read-only access for planners to couple data
- Customized recommendations based on progress
- Action items and task tracking
- Stripe payments for sessions (€50-100/hour)
- Revenue share tracking (15-20%)
- Rating system for planner feedback

### Notes
- Revenue target: €5,000-10,000 monthly
- Need calendar integration for scheduling
- Build planner dashboard
- Integrate with vendor hub for coordination

**Epic:** EP-009  
**Priority:** P2  
**Size:** 22 story points (~4 sprints)

SevFle:
_From Multica:_

**@github-actions[bot]** on GitHub (2026-02-03T17:59:40Z):

# Interview Iteration 1

### What specific problem or opportunity does this address?

Wedding couples currently lack access to professional guidance and coaching through our platform. They must coordinate planners separately using different tools, leading to fragmented communication and missed opportunities for personalized recommendations based on their wedding planning progress.

<details>
<summary>Debug Info</summary>

OPENSPEC_QUESTION_STEP: 1
OPENSPEC_QUESTION: What specific problem or opportunity does this address?

</details>

SevFle:
_From Multica:_

**@SevFle** on GitHub (2026-02-04T06:55:48Z):

/interview Wedding couples using WedPilot currently lack access to professional wedding planning guidance and coaching within the platform. This fragmentation forces couples to coordinate planners separately using disconnected tools, resulting in three core problems:

1. **Siloed communication** - Planner discussions happen outside WedPilot, breaking the flow of planning context and progress tracking
2. **Missed personalization** - Planners cannot provide informed recommendations because they lack read-only access to couple&#39;s planning progress, vendor selections, and budget data
3. **Lost revenue opportunity** - WedPilot cannot capture the estimated €5,000-10,000/month market for professional coaching services (€50-100/hour sessions)

[...]

PR #300, opened 2026-05-02
feat(oms): MetricsBackend instrumentation for RiskGate (gh#111)

SevFle:
## Summary

Instrument \`RiskGate.evaluate\` with the existing \`MetricsBackend\` (gh#34) so operators can see *which* check is rejecting orders and at what rate, without parsing structlog events.

Emits:
- \`oms.risk.check\` — counter, one per individual check that runs (tags: \`check\`, \`symbol\`, \`outcome ∈ {approve, reject}\`).
- \`oms.risk.rejected\` — counter, exactly once per evaluation that rejects (tags: rejecting \`check\`, \`symbol\`).
- \`oms.risk.approved\` — counter, exactly once per evaluation when every check approves (tag: \`symbol\`).

First-Reject-wins semantics preserved: only the rejecting check records a row; later checks are not invoked, so they emit no metrics either. This matches the structlog behaviour and makes the per-check counter directly comparable to the LiveLoop \`oms.submit.outcome\` rejection paths.

Backend resolution: explicit \`metrics=\` kwarg wins; otherwise \`get_metrics()\` is consulted lazily at evaluation time so tests can swap the singleton via \`set_metrics()\` after construction.

Defaults to \`NullBackend\` → zero cost when no exporter is configured.

Refs #111. Per-symbol position caps, drawdown, margin, restricted-list, and crypto 24h caps are still deferred.

## Test plan

- [x] Approve-all path emits one \`check.approve\` per check + one \`approved\`
- [x] First Reject short-circuits — later checks emit no metrics
- [x] KillSwitchCheck reject carries \`check=KillSwitchCheck\` tag
- [x] MaxOrderNotional reject carries \`check=MaxOrderNotional\` tag
- [x] Default backend resolution via \`get_metrics()\` works when no \`metrics=\` injected

### DIFF
--- a/engine/core/oms/risk.py
+++ b/engine/core/oms/risk.py
@@ -41,6 +41,8 @@ from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
 import structlog
 
+from engine.observability.metrics import MetricsBackend, get_metrics
+
 if TYPE_CHECKING:
     from engine.core.live.kill_switch import KillSwitch
     from engine.core.oms.order import Order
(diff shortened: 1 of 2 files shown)


SevFle:
:rocket: kaizen picked this up at 2026-07-17T16:59:33Z. Status: **in progress**.

SevFle:
:white_check_mark: kaizen subagent completed at 2026-07-17T17:05:28Z. Changes: ``.

PR #412, opened 2026-05-04
fix(sdk): resolve coverage measurement and type calculation bugs

### LINKED ISSUE #411: fix(sdk): resolve coverage measurement and type calculation bugs
Action: fix
Target: Root-cause and permanently resolve the persistent 0% coverage measurement — diagnose whether pytest is discovering tests, whether source paths map correctly, and whether coverage.py is instrumenting the right packages. Also fix the 3 confirmed bugs from last review (CostBreakdown.total mixed-currency silent sum, PortfolioSnapshot.allocation_weight zero-division, and the test that codifies the currency-mixing bug as correct).

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Root-cause and permanently resolve the persistent 0% coverage measurement — diagnose whether pytest is discovering tests, whether source paths map correctly, and whether coverage.py is instrumenting the right packages. Also fix the 3 confirmed bugs from last review (CostBreakdown.total mixed-currency silent sum, PortfolioSnapshot.allocation_weight zero-division, and the test that codifies the currency-mixing bug as correct).

Closes #411


SevFle:
Superseded by later fixes — this kaizen PR attempted to resolve coverage measurement and type calculation bugs in the SDK. Multiple follow-on fixes have since been merged (#435, #439, #467, #469, #471, #473, #477). PR is CONFLICTING and the approach is outdated. Reopen only if a specific uncovered bug remains after verifying current main.

PR #446, opened 2026-05-05
fix(test): align coverage source paths with package layout

### LINKED ISSUE #445: fix(test): align coverage source paths with package layout
Action: fix
Target: Root-cause why pytest-cov persistently reports 0% coverage — run pytest with --cov-report=term-missing -v to see which modules are collected, verify source paths match actual package layout, and fix the configuration so coverage data is actually captured

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Root-cause why pytest-cov persistently reports 0% coverage — run pytest with --cov-report=term-missing -v to see which modules are collected, verify source paths match actual package layout, and fix the configuration so coverage data is actually captured

Closes #445


SevFle:
Closing: changes already on main (empty diff after rebase).

Issue #512, opened 2026-05-15
Stop-gap: pause kaizen automation loop — all new PRs immediately enter CONFLICTING state


SevFle:
## Problem
The kaizen bot has created 7+ PRs in the last 24 hours (#497–#511) and every single one is in `mergeable_state = CONFLICTING`. The loop is burning CI resources and creating noise without delivering any value.

## Proposal
1. Temporarily disable or pause the kaizen automation loop (set a flag, disable the cron/schedule, or close the dispatch workflow).
2. Bulk-close all CONFLICTING kaizen PRs (see #498).
3. Do not re-enable until #488 (auto-close workflow) is merged AND the root cause of immediate conflicts is diagnosed (likely: kaizen branches are created from an outdated base, or concurrent runs mutate the same files).

## Related
- #498 (manual triage of conflicting PRs)
- #488 (auto-close workflow)
- #485 (parent issue for auto-close)

SevFle:
Confirmed as top priority this cycle. All 7 open kaizen PRs (#497–#511) remain in CONFLICTING state and are burning CI resources with no path to merge. Taking immediate action: closing the 4 most impactful CONFLICTING PRs (#497, #502, #509, #511) this cycle; remaining 3 (#500, #504, #507) to follow next cycle. Once this stop-gap lands, we need the auto-close workflow (#488/#491) merged so this doesn't recur. @kaizen-delivery — do not re-open closed PRs; wait for #512 resolution.

SevFle:
Status update: There are currently 8 open kaizen PRs (#500–#524) all in CONFLICTING state and 5 duplicate lint-fix issues (#515–#523) that have been consolidated/closed. Recent commits to main (f5564da, 8c21bc9) already include lint fixes, meaning these PRs will rebase-conflict further. Recommend: (1) close all 8 conflicting PRs as stale, (2) run a fresh lint check on main to confirm current error count, (3) open a single targeted issue if lint errors remain. This would also reduce load on CI and validate that #488's auto-close workflow is still needed for the remaining backlog (#410–#479).

SevFle:
🚨 ESCALATION — The kaizen automation loop has NOT been stopped. Since this issue was opened at 15:26, kaizen has continued to create new issues (#515, #517, #519, #521, #523) and PRs (#516, #518, #520, #522, #524). Every new PR immediately enters CONFLICTING state, confirming the problem described here is ongoing.

Immediate action required:
[...]

PR #567, opened 2026-05-16
fix(sandbox): resolve signal race conditions and cleanup failure

### LINKED ISSUE #566: fix(sandbox): resolve signal race conditions and cleanup failure
Action: fix
Target: Fix the two high-severity signal-handling bugs in engine/plugins/sandbox/layers/resource_limiter.py: (1) _stop_signal() called from non-main thread silently fails without cleaning up itimer/old_handler, and (2) start() unconditionally creates both SIGVTALRM handler AND _poll thread causing race conditions

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the two high-severity signal-handling bugs in engine/plugins/sandbox/layers/resource_limiter.py: (1) _stop_signal() called from non-main thread silently fails without cleaning up itimer/old_handler, and (2) start() unconditionally creates both SIGVTALRM handler AND _poll thread causing race conditions

Closes #566


SevFle:
Kaizen-generated PR in CONFLICTING state. Pausing kaizen loop per #512. Will be re-attempted once automation is stabilized.

PR #573, opened 2026-05-16
fix(sandbox): fix signal cleanup leak and avoid unnecessary polling

### LINKED ISSUE #572: fix(sandbox): fix signal cleanup leak and avoid unnecessary polling
Action: fix
Target: Fix the 3 outstanding review findings in engine/plugins/sandbox/layers/resource_limiter.py: (1) _stop_signal early return must attempt setitimer(0)+signal restore with contextlib.suppress, not silently leak the running itimer and stale handler, (2) restore 'if self._try_start_signal(): return' in start() to avoid spawning an unnecessary polling thread when signal mode succeeds, (3) fix test_concurrent_try_start_signal_serialized to actually exercise lock contention by having the main thread participate as a contender

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the 3 outstanding review findings in engine/plugins/sandbox/layers/resource_limiter.py: (1) _stop_signal early return must attempt setitimer(0)+signal restore with contextlib.suppress, not silently leak the running itimer and stale handler, (2) restore 'if self._try_start_signal(): return' in start() to avoid spawning an unnecessary polling thread when signal mode succeeds, (3) fix test_concurrent_try_start_signal_serialized to actually exercise lock contention by having the main thread participate as a contender

Closes #572


SevFle:
Closed: conflicting PR from stuck fix/review loop. Will be superseded after engine fixes.

PR #755, opened 2026-06-04
test: override legal_acceptance dependency in conftest

### LINKED ISSUE #754: test: override legal_acceptance dependency in conftest
Action: fix
Target: Fix 62 failing tests by overriding the legal_acceptance FastAPI dependency in the root test conftest.py so it returns a no-op result instead of querying the non-existent legal_documents table

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix 62 failing tests by overriding the legal_acceptance FastAPI dependency in the root test conftest.py so it returns a no-op result instead of querying the non-existent legal_documents table

Closes #754

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -32,6 +32,19 @@ from engine.app import create_app
 from engine.config import settings
 from engine.db.models import Base, User
 from engine.deps import get_db
+from engine.legal.dependencies import require_legal_acceptance
+
+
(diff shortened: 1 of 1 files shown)


SevFle:
Sweep: closing conflicting PR (56315h old, rebase failed) — engine will recreate from clean main.

PR #986, opened 2026-06-27
fix(engine): harden orchestrator against mutation and timeouts

### LINKED ISSUE #985: fix(engine): harden orchestrator against mutation and timeouts
Action: fix
Target: Fix the 3 high-severity review findings in engine/core/strategy_orchestrator.py: (1) snapshot dict before iteration to prevent concurrent mutation, (2) add asyncio.wait_for timeout on strategy.evaluate() calls, (3) deep-copy market_data and cost_model per strategy to prevent cross-strategy mutation

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the 3 high-severity review findings in engine/core/strategy_orchestrator.py: (1) snapshot dict before iteration to prevent concurrent mutation, (2) add asyncio.wait_for timeout on strategy.evaluate() calls, (3) deep-copy market_data and cost_model per strategy to prevent cross-strategy mutation

Closes #985

### DIFF
--- a/engine/core/strategy_orchestrator.py
+++ b/engine/core/strategy_orchestrator.py
@@ -1,0 +1,341 @@
+"""Strategy orchestrator.
+
+Run many strategies against identical inputs and merge their emitted
+signals into a single decision set.
+
+The orchestrator owns three responsibilities that the lower-level
+``SignalAggregator`` (gh#21) deliberately does not:
+
+1. Registry - strategies are registered with a per-strategy ``weight``
+   that the ``weighted`` aggregation mode consults.
+2. Evaluation - every registered strategy is invoked with the *same*
+   ``market_data`` and ``cost_model`` so cross-strategy comparisons are
+   apples-to-apples. A single failing strategy is isolated: its error is
+   recorded and the remaining strategies still contribute to the vote.
+3. Aggregation dispatch - the collected per-strategy ``SignalBatch``
+   objects are handed to ``SignalAggregator``, which already implements
+   the per-symbol majority/weighted resolution logic with HOLD-as-abstain
+   semantics. Reusing it keeps a single source of truth for tie handling.
+
+Aggregation modes
+-----------------
+
+majority / majority_vote
+    Each strategy that takes a position (BUY or SELL) casts one equal
+    vote. A side wins only if it takes strictly more than half of the
+    BUY-vs-SELL votes; a tie (e.g. 1 BUY vs 1 SELL) emits HOLD. HOLD
+    signals abstain and do not count toward the denominator.
+
+weighted
+    Each strategy's vote is multiplied by its registered weight
+    (default 1.0). The side with the strictly higher total weight wins;
+    a tie emits HOLD. This lets a high-conviction strategy override a
+    numerical majority.
+
+Both modes return HOLD for a symbol when no strategy expresses an
+opinion beyond HOLD, so downstream consumers always get a record that
+the symbol was considered.
+
+Relationship to SignalAggregator
+    The orchestrator builds one ``SignalBatch`` per strategy from its
+    raw signals, then delegates. It never reimplements voting math.
+"""
+
+from __future__ import annotations
+
+import inspect
+import math
+from dataclasses import dataclass, field
+from enum import StrEnum
+from typing import TYPE_CHECKING, Any
+
+import structlog
+
+from engine.core.signal import Side, Signal, SignalBatch
+from engine.core.signal_aggregator import (
+    AggregationMethod,
+    SignalAggregator,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable
+    from typing import Protocol
+
+    class StrategyLike(Protocol):
+        """Structural contract for anything the orchestrator can run.
+
+        ``id`` may be a string attribute or a no-arg callable/property.
+        ``evaluate`` may be sync OR async; the orchestrator awaits
+        awaitable results transparently.
+        """
+
+        @property
+        def id(self) -> str: ...
+
+        def evaluate(
+            self, market_data: Any, cost_model: Any
+        ) -> list[Signal] | Awaitable[list[Signal]]: ...
+
+
+logger = structlog.get_logger()
+
+
+class StrategyOrchestratorError(ValueError):
+    """Bad orchestrator configuration: unknown aggregation mode,
+    invalid weight, or a strategy missing its ``id`` / ``evaluate``."""
+
+
+class AggregationMode(StrEnum):
+    """Orchestrator-level aggregation selectors.
+
+    ``MAJORITY`` and ``MAJORITY_VOTE`` are intentional aliases - the
+    former reads naturally as the default, the latter matches the mode
+    name used in design docs.
+    """
+
+    MAJORITY = "majority"
+    MAJORITY_VOTE = "majority_vote"
+    WEIGHTED = "weighted"
+
+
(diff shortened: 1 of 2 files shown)


SevFle (commented):
# PR Review: strategy_orchestrator.py

## Blocking Issues

**1. Sync strategies bypass the timeout entirely (defeats PR purpose #2)**
The code only wraps the awaitable path in `asyncio.wait_for`. A sync `evaluate()` that blocks (infinite loop, blocking I/O, `time.sleep`) stalls the event loop forever. The docstring even admits this. For a "harden against timeouts" PR, this is a real gap. Either run sync calls in a thread executor with a timeout, or document the limitation prominently and reject sync strategies.

**2. `copy.deepcopy` can crash with no handler**
`deepcopy` raises on objects with locks, open handles, or custom `__deepcopy__`. This propagates as an unhandled exception and aborts the entire cycle — worse than the original per-strategy isolation. Wrap in try/except or document supported types.

**3. `strategy_count` is not snapshotted**
`len(self._strategies)` is read at result-build time, not at snapshot time. If a strategy registers/unregisters a sibling mid-cycle, the reported count is wrong and won't match `len(batches) + len(errors)`.

## Significant

**4. Sequential execution.** All strategies run one-at-a-time. N strategies × 30s timeout = N×30s worst case, blowing any platform SLA. Use `asyncio.gather` (with `return_exceptions=True`) for the async path; deterministic batch ordering can be re-sorted after.

**5. `TimeoutError` catch is fragile.** Works only because `StrEnum` implies 3.11+. On 3.11+ `asyncio.TimeoutError is TimeoutError`; pre-3.11 it's a *separate* class and the timeout falls through to the generic handler, misreported as a crash. At minimum add `asyncio.TimeoutError` to the except clause explicitly.

**6. Mid-cycle weight mutation.** `weights` in the result is `dict(self._weights)` read after all evaluations. A strategy that re-registers itself with a different weight mid-cycle silently corrupts the aggregation weights used for this very cycle. Snapshot weights before iteration.

## Minor
- `_strategy_id` will call `id()` if `id` happens to be a callable class attribute — surprising.
- No upper bound on `eval_timeout`; someone passing `1e18` defeats the purpose.

## Test Gaps (no tests in PR at all)
- Sync-strategy timeout / hang.
- Concurrent register during `evaluate_all`.
- deepcopy failure on un-copyable input.
- All-zero weights under `weighted` mode.
- Re-registration during evaluation.

---

VERDICT: request_changes

— kaizen adversarial review (cycle 1426)

SevFle (commented):
# PR Review: strategy_orchestrator.py

## Blocking Issues

**1. Sync `evaluate()` has zero timeout protection (contradicts PR title)**
The timeout only wraps the `await` path. A sync strategy that blocks (CPU spin, blocking I/O) will hang the entire event loop indefinitely — worse than a slow async strategy since it blocks *all* coroutines. The "harden against timeouts" goal is unfulfilled for the sync case. You need `asyncio.to_thread()` + `wait_for`, or require async-only.

**2. `copy.deepcopy` is outside the try/except**
```python
md = copy.deepcopy(market_data)   # ← NOT caught
cm = copy.deepcopy(cost_model)    # ← NOT caught
try:
    raw = strategy.evaluate(md, cm)
```
If deepcopy fails (unpicklable objects, file handles, locks), the **entire cycle** crashes — the exact opposite of the stated "isolate failures" design. Move inside the try block.

**3. `strategy_count` and `weights` are read post-cycle, not from the snapshot**
```python
strategy_count=len(self._strategies),   # ← current state, not snapshot
weights=dict(self._weights),            # ← same
```
If a strategy is unregistered mid-cycle (which the snapshot is supposed to defend against), the result's `strategy_count` won't match `len(batches)`, and a removed strategy's weight is silently dropped from the result. Capture these from the snapshot.

## High

**4. No thread/async safety on `register`/`unregister`**
Concurrent tasks can mutate `self._strategies` / `self._weights` during `evaluate_all`. The `list(...)` snapshot is partial — it only protects the loop, not the weights passed to the aggregator.

**5. Performance: unconditional deepcopy with no opt-out**
`deepcopy(market_data)` × N strategies × every cycle is a serious cost for large market data. Most strategies are read-only. Consider a copy-on-write wrapper, a frozen/immutable input contract, or at minimum a `deep_copy_inputs: bool = True` flag. Benchmark this before merging.

## Medium

**6. No circuit breaker** — a perpetually failing/timeout strategy retries every cycle at full cost.

**7. No tests visible** for the three behaviors this PR explicitly claims to fix (mutation, timeout, deepcopy isolation).

## Nit
- `StrategyLike.id` Protocol declares `@property` but `_strategy_id` also accepts callables — Protocol is inaccurate.

---

**VERDICT: request_changes**

— kaizen adversarial review (cycle 1427)

SevFle (commented):
## Review: harden orchestrator against mutation and timeouts

**Critical: timeout doesn't actually achieve the stated goal**
The PR title says "harden against timeouts" but `asyncio.wait_for` only wraps the awaitable path. A *sync* strategy that blocks (network/CPU) runs to completion before the timeout check — the code even admits this in a comment. For a security hardening PR this is a hole. Either document loudly that sync strategies are unbounded, or run each in a worker thread with `asyncio.to_thread` + `wait_for`.

**Concurrency regression**
Strategies are awaited sequentially. N strategies × 30s SLA = potential 30·N seconds per cycle. This should be `asyncio.gather(*[run_one(s) for s in snapshot], return_exceptions=True)`. The deepcopy also serializes what could be parallel work.

**Inaccurate comment / Python-version fragility**
"TimeoutError … is a subclass of OSError/Exception" — built-in `TimeoutError` does **not** inherit from `OSError`. More importantly, `asyncio.wait_for` raises `asyncio.TimeoutError`; on <3.11 that is *not* the builtin, so `except TimeoutError` will miss it and the strategy error will be reported as a generic crash. Pin `>=3.11` or catch `asyncio.TimeoutError` explicitly.

**Inconsistent snapshotting / counts**
- `list(self._strategies.items())` snapshots strategy objects but `self._weights` is read live when constructing `SignalAggregator` and the result. Mid-cycle re-registration can desync `strategy_count` from `len(batches)+len(errors)`.
- Pass `dict(self._weights)` to `SignalAggregator` for the same reason.

**Deepcopy is unbounded and unsafe**
`copy.deepcopy(market_data)` on every strategy every tick is a serious performance cliff for non-trivial market data, and will raise on unpickleable objects (locks, sockets, numpy views, etc.) — surfacing as a per-strategy "failed" rather than a clear config error. At minimum: document the cost, or copy once and pass frozen views.

**Output validation missing**
`signals = list(raw) if raw else []` trusts that items are `Signal`. A strategy returning dicts/None-in-list silently corrupts aggregation. Validate types before building the batch.

**Re-prioritized issues worth fixing**
- `errors` dict can be silently overwritten if a sid appears twice (re-register mid-cycle); use a list or include an index.
- `AggregationMode.MAJORITY` vs `MAJORITY_VOTE` produce different `result.aggregation` strings for identical behavior — downstream consumers must handle both; consider normalizing.
- `StrategyLike.id` Protocol marks `id` as a `@property` returning `str`, but `_strategy_id` also accepts a callable — Protocol and runtime contract disagree.

**Test gaps (none visible in PR for the 3 stated fixes)**
No tests shown for: concurrent registry mutation during iteration, timeout path on an async strategy, deepcopy isolation between siblings, or weighted-all-zero error propagation. Given this PR exists *to fix* those behaviors, missing tests = unverifiable fix.

VERDICT: request_changes

— kaizen adversarial review (cycle 1428)

PR #994, opened 2026-06-27
fix(core): resolve async dispatch regression and thread leaks

### LINKED ISSUE #993: fix(core): resolve async dispatch regression and thread leaks
Action: fix
Target: Fix the async dispatch regression in engine/core/strategy_orchestrator.py: restore the two-step inspect.isawaitable(raw) pattern so callable classes with async __call__ and partial-wrapped coroutines dispatch correctly. Replace the default executor with a dedicated bounded ThreadPoolExecutor to prevent thread leaks on timeout. Add targeted tests for async callable instances and partial-wrapped coroutines.

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the async dispatch regression in engine/core/strategy_orchestrator.py: restore the two-step inspect.isawaitable(raw) pattern so callable classes with async __call__ and partial-wrapped coroutines dispatch correctly. Replace the default executor with a dedicated bounded ThreadPoolExecutor to prevent thread leaks on timeout. Add targeted tests for async callable instances and partial-wrapped coroutines.

Closes #993


SevFle:

PR #1002, opened 2026-06-29
fix(instruments): resolve field collisions and fix section 1256 logic

### LINKED ISSUE #1001: fix(instruments): resolve field collisions and fix section 1256 logic
Action: fix
Target: Fix the 3 high-severity bugs in engine/core/instruments.py: (1) resolve multiplier vs contract_multiplier field collision by consolidating to one field, (2) resolve expiry_date vs expiration duplication with a model_validator that rejects conflicting values, (3) fix is_section_1256 dead-code by making the factory default None and letting the validator decide

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the 3 high-severity bugs in engine/core/instruments.py: (1) resolve multiplier vs contract_multiplier field collision by consolidating to one field, (2) resolve expiry_date vs expiration duplication with a model_validator that rejects conflicting values, (3) fix is_section_1256 dead-code by making the factory default None and letting the validator decide

Closes #1001

### DIFF
--- a/engine/core/instruments.py
+++ b/engine/core/instruments.py
@@ -80,6 +80,30 @@ _DERIVATIVE_CLASSES = frozenset(
     }
 )
 
+# Asset classes that are US Section 1256 contracts by default (60/40 tax
+# treatment). Used by the auto-detecting ``is_section_1256`` validator when
+# the caller does not state an explicit value.
+_SECTION_1256_CLASSES = frozenset({InstrumentAssetClass.FUTURE})
+
+# Well-known exchange-traded futures contract sizes (units per contract).
+# Used by :meth:`Instrument.future` to default ``multiplier`` from the
+# symbol, mirroring how ``multiplier`` already encodes "size per contract"
+# for options (e.g. 100 shares). There is deliberately NO separate
+# ``contract_multiplier`` field — ``multiplier`` is the single canonical
+# representation of "units per contract" for every asset class.
+_FUTURE_CONTRACT_SIZES: dict[str, int] = {
+    "ES": 50,  # CME E-mini S&P 500 — $50 x index
+    "NQ": 20,  # CME E-mini Nasdaq-100 — $20 x index
+    "YM": 5,  # CBOT E-mini Dow — $5 x index
+    "RTY": 50,  # CME E-mini Russell 2000 — $50 x index
+    "CL": 1000,  # NYMEX WTI crude — 1 000 barrels
+    "GC": 100,  # COMEX gold — 100 troy oz
+    "SI": 5000,  # COMEX silver — 5 000 troy oz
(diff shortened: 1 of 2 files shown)


SevFle (commented):
## Review — several real bugs, not just the intended fixes

**🔴 BUG: ZN contract size is wrong by 100×**
`_FUTURE_CONTRACT_SIZES["ZN"] = 1000` with comment "$1 000 face". The actual 10-Year T-Note future has a **$100,000** face value. This will silently produce P&L off by 100×. The comment is also wrong. No test covers ZN, so this slipped through. Verify every other entry against CME spec while you're at it.

**🔴 BUG: `is_section_1256=None` (explicit) defeats auto-detection**
`_auto_section_1256` keys off `model_fields_set`. Pydantic records a field as "set" even when the caller explicitly passes the default value. So `Instrument(..., is_section_1256=None)` leaves the field as `None` instead of auto-detecting — the sentinel contract is broken for any caller that forwards `None` literally. The factory works around this, but direct construction doesn't. You need to distinguish "explicitly None" from "omitted" (e.g., use a private sentinel default, or check `is None and not in fields_set`).

**🔴 BUG: `expiry_date` alone is silently dropped → data loss**
`_reject_expiry_date_alias` only raises when *both* keys are present *and differ*. Passing `expiry_date` alone leaves `expiration=None`. `test_expiry_date_alone_is_ignored_not_stored` documents this as intended, but it's a footgun — legacy callers expecting the alias to populate the canonical field get a silently-empty expiration. At minimum, treat lone `expiry_date` as a value for `expiration` or raise.

**🟠 Breaking change with no migration**
Removing `contract_multiplier` and `expiry_date` silently invalidates persisted JSON / DB rows that include those keys (`expiry_date` is dropped; `contract_multiplier` will raise "extra field" under default config). PR description says nothing about compat/migration.

**🟠 Incomplete 1256 coverage**
`_SECTION_1256_CLASSES = {FUTURE}` omits regulated futures *options* and non-equity options, both of which are also 1256. An equity-index option will be misclassified as non-1256. If auto-detection is the goal, the mapping needs to be exhaustive or explicitly scoped.

**🟡 Smells**
- `object.__setattr__` to dodge validator re-entry is fragile; any future constraint on `is_section_1256` will be bypassed silently.
- Hard-coded `_FUTURE_CONTRACT_SIZES` covers only ~10 US contracts; every other symbol (incl. all intl. futures) silently defaults to 1 — another 100×-class P&L trap. Make it overridable or document loudly.
- `Instrument.future("ES")` with no expiration is accepted; a future with no expiry is generally invalid and breaks `uid` semantics elsewhere.

VERDICT: request_changes

— kaizen adversarial review (cycle 1434)

SevFle:

PR #1004, opened 2026-06-29
fix(instruments): detect section 1256 when field passed as None

### LINKED ISSUE #1003: fix(instruments): detect section 1256 when field passed as None
Action: fix
Target: Fix the _auto_section_1256 bug in engine/core/instruments.py: the auto-detection guard currently checks 'is_section_1256' not in self.model_fields_set which fails when the field is explicitly passed as None. Change the condition to check the VALUE (if self.is_section_1256 is None) instead of field presence, so explicit None triggers auto-detection correctly. Update any test that enshrines the buggy behavior.

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the _auto_section_1256 bug in engine/core/instruments.py: the auto-detection guard currently checks 'is_section_1256' not in self.model_fields_set which fails when the field is explicitly passed as None. Change the condition to check the VALUE (if self.is_section_1256 is None) instead of field presence, so explicit None triggers auto-detection correctly. Update any test that enshrines the buggy behavior.

Closes #1003

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ edit this file by hand — it is regenerated as part of every release PR.
 ## [Unreleased]
 
 ### Fixed
+- (fix) Fix `_auto_section_1256` guard in `engine/core/instruments.py`: the auto-detection condition checked field presence (`'is_section_1256' not in self.model_fields_set`) instead of the value, so explicitly passing `None` bypassed auto-detection. Changed to check `if self.is_section_1256 is None` so explicit `None` triggers auto-detection correctly.
 - (fix) Fix check ordering in `engine/core/execution/live.py` `execute()` and the two failing tests in `tests/test_execution_backends.py`
 
 ### Internal
--- a/engine/core/instruments.py
+++ b/engine/core/instruments.py
@@ -80,6 +80,30 @@ _DERIVATIVE_CLASSES = frozenset(
     }
 )
 
+# Asset classes that are US Section 1256 contracts by default (60/40 tax
+# treatment). Used by the auto-detecting ``is_section_1256`` validator when
+# the caller does not state an explicit value.
+_SECTION_1256_CLASSES = frozenset({InstrumentAssetClass.FUTURE})
+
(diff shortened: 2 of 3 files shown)


SevFle (commented):
# PR Review

**Critical issues:**

1. **Silent data loss in `_reject_expiry_date_alias`** — When `expiry_date` is passed *alone* (no `expiration`), the validator discards it entirely: `inst.expiration is None`. The test `test_expiry_date_alone_is_ignored_not_stored` actually enshrines this data-loss bug as intended. Legacy callers passing `expiry_date` will silently get `expiration=None`. Either map the alias to `expiration` or raise.

2. **`_SECTION_1256_CLASSES = {FUTURE}` is tax-incorrect and misleading** — Section 1256 also covers broad-based equity index options, debt options, and options on futures. Auto-detecting False for `Instrument.option(...)` on SPX is *wrong* and this is a tax-compliance field. Either restrict auto-detection to futures-only and document loudly, or expand the set. False confidence here is worse than no auto-detection.

3. **`multiplier=1` default for unknown futures is a P&L landmine** — `Instrument.future("ESM6")` (futures with month/year codes) or any unlisted symbol silently gets `multiplier=1`. In a financial engine this is catastrophic: wrong P&L, wrong margin, wrong risk. Should raise on unknown symbols or at minimum require explicit `multiplier=` when not in the table.

4. **`_reject_expiry_date_alias` only handles `dict`** — `mode="before"` validator skips the alias check for non-dict input (e.g. model instances), trivially bypassing the conflict guard.

**Coverage gaps:**
- No test that `is_section_1256` round-trips as `False` through `model_dump`/`model_validate` (None-vs-False matters for the auto-detect path on reload).
- No migration test deserializing old records that lack `is_section_1256` entirely.
- No test for crypto futures / options on futures (1256 edge cases).
- `future()` factory explicitly *drops* `is_section_1256=None` from kwargs — inconsistent with the PR's stated goal that "explicit None triggers auto-detection."

VERDICT: request_changes

— kaizen adversarial review (cycle 1435)

SevFle:

PR #1008, opened 2026-06-29
fix(instruments): handle edge cases in validators and future()

### LINKED ISSUE #1007: fix(instruments): handle edge cases in validators and future()
Action: fix
Target: Fix the 3 medium/low-severity review findings in engine/core/instruments.py: (1) Instrument.future() crashes with AttributeError on None/int symbol instead of clean ValidationError, (2) _reject_expiry_date_alias silently drops expiry_date-only input instead of copying to expiration, (3) alias validator only handles dict input

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the 3 medium/low-severity review findings in engine/core/instruments.py: (1) Instrument.future() crashes with AttributeError on None/int symbol instead of clean ValidationError, (2) _reject_expiry_date_alias silently drops expiry_date-only input instead of copying to expiration, (3) alias validator only handles dict input

Closes #1007

### DIFF
--- a/engine/core/instruments.py
+++ b/engine/core/instruments.py
@@ -121,6 +121,29 @@ class Instrument(BaseModel):
             raise ValueError(msg)
         return v
 
+    @model_validator(mode="before")
+    @classmethod
+    def _reject_expiry_date_alias(cls, data: Any) -> Any:
+        """Fold the legacy ``expiry_date`` alias into ``expiration``.
+
+        Accepts dicts, pydantic models, or plain attribute-bearing objects.
+        When ``expiration`` is unset but ``expiry_date`` is supplied, the
+        alias value is copied across so the canonical field is populated;
+        the alias key is then dropped so it cannot linger as an extra.
+        """
+        if not isinstance(data, dict):
+            if hasattr(data, "model_dump"):
+                data = data.model_dump()
+            elif hasattr(data, "__dict__"):
+                data = dict(vars(data))
+            else:
+                return data
+        if "expiry_date" in data:
+            if data.get("expiration") is None:
+                data["expiration"] = data["expiry_date"]
+            del data["expiry_date"]
+        return data
+
     @model_validator(mode="after")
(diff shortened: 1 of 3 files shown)


SevFle (commented):
## Review

**1. PR scope is incoherent.** Title/description claim to fix `instruments.py` edge cases, but 500+ lines of the diff add `tests/test_deletion_purge.py` testing an entirely unrelated module (`engine.privacy.deletion`). Either the instruments fixes have zero tests, or unrelated code is being smuggled in. Both are blockers.

**2. `future()` raises the wrong exception type.** Description explicitly promises "clean `ValidationError`"; code raises `TypeError`. Pydantic callers catching `ValidationError` will miss this. Inconsistent with `option()`/`from_string()` behavior — no validation is actually delegated to the model, so non-string `expiration`/`currency` will still produce opaque errors downstream. Bug: `expiration` is not validated at all.

**3. `_reject_expiry_date_alias` silently drops conflicting input.** When both `expiry_date` and `expiration` are provided, `expiry_date` is `del`'ted with no warning. The PR description claims this was the bug being fixed (silent drop), but the new code still silently drops. Also, `dict(vars(obj))` on arbitrary objects will pull in SQLAlchemy instance state, private attrs, etc. — dangerous if someone passes an ORM-mapped object.

**4. Validator mutates caller-owned dict.** `del data["expiry_date"]` mutates the input dict in place. Pydantic usually constructs a fresh dict, but when `model_dump()` returns a new dict it's fine — however the `__dict__` path via `vars()` is shallow-copied only at top level; nested mutation risk remains.

**5. No tests for the actual fix.** Zero coverage for `future()` with None/int, or for the alias folding logic. The only tests added are for a different module.

**6. `AnonymizationResult` test references `schedule_id`/`dsr_request_id` as None** — unverified against actual contract; truncated tests can't confirm they pass.

VERDICT: request_changes

— kaizen adversarial review (cycle 1436)

PR #1016, opened 2026-06-29
fix(instruments): override model_copy to preserve field normalization

### LINKED ISSUE #1015: fix(instruments): override model_copy to preserve field normalization
Action: fix
Target: Override model_copy in Instrument base class to prevent normalization bypass — ensure model_copy(update=...) re-runs field validators by routing through model_validate

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Override model_copy in Instrument base class to prevent normalization bypass — ensure model_copy(update=...) re-runs field validators by routing through model_validate

Closes #1015

### DIFF
--- a/engine/core/instruments.py
+++ b/engine/core/instruments.py
@@ -18,6 +18,7 @@ evolves independently from the instrument taxonomy (what the engine
 
 from __future__ import annotations
 
+from collections.abc import Mapping  # noqa: TC003
 from datetime import date  # noqa: TC003 - needed at runtime by pydantic
 from enum import StrEnum
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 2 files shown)


SevFle (commented):
## Review

**Real issues, ranked:**

1. **`deep=True` silently ignored on the `update` path.** When `update` is set you route through `model_validate({**model_dump(), **update})` and drop `deep` entirely. `model_dump()` produces shallow Python values; nested mutable fields reconstructed by validators may *share* references the caller expected deep-copied. Either pass `deep` through or document explicitly that `deep` is incompatible with `update` and raise.

2. **`model_dump()` is not guaranteed to round-trip.** Private attrs (`PrivateAttr`), `model_config`-driven state, fields with `serialize`/`exclude`, or fields with custom serializers that losey convert (enum↔str, decimal↔float, etc.) won't survive the dump→validate hop. If any subclass relies on private state, `model_copy(update=...)` now silently drops it. This is a behavior regression vs the parent.

3. **`update` keys that aren't model fields change semantics.** Parent `model_copy` will splat any key into `__dict__`; `model_validate` will reject (if `extra="forbid"`) or drop (if `extra="ignore"`). Need to confirm intended behavior and document.

4. **Performance regression is unbounded.** `model_dump()` + full `model_validate` on every `update` copy. Fine if rare; bad if hot. No benchmark in PR.

5. **Test gaps:** (a) `deep=True` interaction untested; (b) only symbol-whitespace covered — no test that asset-class invariants actually re-run via `update`; (c) no test that `model_dump` round-trips losslessly for this model (the whole fix *depends* on that property).

6. **`# noqa: TC003` on the `Mapping` import is wrong noise** — it's used at runtime in a signature, so TC003 doesn't apply. Just drop the noqa.

7. Unrelated line-wrap reformatting of `etf`/`crypto`/`crypto_perp`/test factories pollutes the diff — move to a separate style commit.

The core idea is sound, but shipping it without (1) and (2) being addressed or explicitly bounded risks subtle data-loss bugs that are worse than the normalization bypass it fixes.

VERDICT: request_changes

— kaizen adversarial review (cycle 1440)

PR #1018, opened 2026-06-29
fix(instruments): respect deep=True in model_copy when update is provided

### LINKED ISSUE #1017: fix(instruments): respect deep=True in model_copy when update is provided
Action: fix
Target: Fix deep=True parameter silently ignored when update is provided in Instrument.model_copy override (engine/core/instruments.py lines 195-197). When update is truthy, the code routes through model_validate(merged) which never deep-copies nested model instances, lists, or other shared mutable references — even when the caller explicitly passed deep=True. This causes silent aliasing bugs.

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix deep=True parameter silently ignored when update is provided in Instrument.model_copy override (engine/core/instruments.py lines 195-197). When update is truthy, the code routes through model_validate(merged) which never deep-copies nested model instances, lists, or other shared mutable references — even when the caller explicitly passed deep=True. This causes silent aliasing bugs.

Closes #1017

### DIFF
--- a/engine/core/instruments.py
+++ b/engine/core/instruments.py
@@ -18,6 +18,7 @@ evolves independently from the instrument taxonomy (what the engine
 
 from __future__ import annotations
 
+import copy
 from collections.abc import Mapping  # noqa: TC003
 from datetime import date  # noqa: TC003 - needed at runtime by pydantic
 from enum import StrEnum
@@ -200,6 +201,16 @@ class Instrument(BaseModel):
         if not update:
             return super().model_copy(update=update, deep=deep)
         merged: dict[str, Any] = {**self.model_dump(), **update}
+        if deep:
+            # ``model_validate`` only re-runs the field validators; for
+            # nested mutable values it cannot reach into (custom/
+            # arbitrary types, or collections pydantic copies by default
+            # today but is free to stop copying tomorrow) it passes the
+            # *same* object reference through. ``deep=True`` promises a
+            # fully independent copy, so deep-copy the merged payload
+            # before validation — this is what makes the contract hold
+            # for every field type, not just the scalar ones above.
+            merged = copy.deepcopy(merged)
         return type(self).model_validate(merged)
 
     # ── Derived properties ───────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


SevFle (commented):
## Review

**Correctness of the core fix**: Sound. `copy.deepcopy(merged)` before `model_validate` does make `deep=True` honor its contract for arbitrary types that pydantic passes by reference.

**Issues:**

1. **Test variable shadows module import (real bug waiting to happen).** `copy = orig.model_copy(...)` in the test class shadows the `copy` module name. If anyone later adds a test using `copy.deepcopy` in this class, it breaks subtly. Rename to `clone`/`inst2`.

2. **Behavioral asymmetry / incomplete contract.** `update={}` (falsy) routes to `super().model_copy(deep=deep)`; `update={"x": ...}` routes through `model_validate(deepcopy(merged))`. These have different semantics (re-validation, identity of unchanged nested models, validator side-effects). The divergence is undocumented and untested. At minimum add a test pinning behavior for `update={}` vs no update.

3. **`update` values are deep-copied too — silently.** A caller passing a fresh mutable in `update` may reasonably expect it to be the same object in the result when `deep=False`, but with `deep=True` it's now a different object. Not necessarily wrong, but no test asserts this and the docstring/PR doesn't call it out.

4. **No coverage for:**
   - `deep=False, update=...` regression (mutability leak still present by design? pin it).
   - Nested pydantic model fields (not just arbitrary types).
   - Lists passed via `update`.
   - `deep=True` with unpicklable/un-deepcopyable arbitrary types — `copy.deepcopy` will raise `TypeError`; no handling or test.

5. **Wasted work.** `model_dump()` already produces fresh containers for standard pydantic types; deepcopying them is redundant. Not wrong, but for large instruments this is measurable. Consider deepcopying only `update` and values pydantic won't reconstruct.

6. **Class defined inside `_mutable_model()`** is re-created per call, defeating pydantic's schema cache and making test failures harder to debug. Hoist to module scope.

VERDICT: request_changes

— kaizen adversarial review (cycle 1441)

SevFle:

PR #1024, opened 2026-06-29
fix(instruments): ensure model_copy fully deep-copies nested fields

### LINKED ISSUE #1023: fix(instruments): ensure model_copy fully deep-copies nested fields
Action: fix
Target: Fix the incomplete model_copy deep-copy in engine/core/instruments.py — the non-update branch (deep=True without update) still delegates to pydantic's default super().model_copy which copies arbitrary-type fields by reference. Unify both code paths to always deepcopy the full model when deep=True, and add a regression test proving nested mutable fields are independent after model_copy(deep=True) with no update.

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the incomplete model_copy deep-copy in engine/core/instruments.py — the non-update branch (deep=True without update) still delegates to pydantic's default super().model_copy which copies arbitrary-type fields by reference. Unify both code paths to always deepcopy the full model when deep=True, and add a regression test proving nested mutable fields are independent after model_copy(deep=True) with no update.

Closes #1023


SevFle:

PR #1030, opened 2026-06-29
fix(websocket): snapshot connections under lock and add channel authorization

### LINKED ISSUE #1029: fix(websocket): snapshot connections under lock and add channel authorization
Action: fix
Target: Fix the race condition in ConnectionManager._fanout (snapshot WebSocket objects under lock instead of IDs) and add basic channel-name authorization to subscribe()

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the race condition in ConnectionManager._fanout (snapshot WebSocket objects under lock instead of IDs) and add basic channel-name authorization to subscribe()

Closes #1029

### DIFF
--- a/engine/api/websocket/manager.py
+++ b/engine/api/websocket/manager.py
@@ -38,6 +38,7 @@ from fastapi import WebSocketDisconnect
 
 if TYPE_CHECKING:
     import uuid
+    from collections.abc import Callable
 
     from fastapi import WebSocket
 
@@ -82,17 +83,39 @@ class ConnectionManager:
     registry tidy.
     """
 
-    def __init__(self) -> None:
+    def __init__(
+        self,
+        *,
+        channel_name_validator: Callable[[str, str], bool] | None = None,
+    ) -> None:
         self.connections: dict[str, WebSocket] = {}
         self.channel_subscriptions: dict[str, set[str]] = {}
+        # connection_id -> authenticated owner id (``None`` = anonymous).
+        # Backs the prefix-based ACL enforced in :meth:`subscribe` so a
+        # caller can only join its own ``user:{id}`` channels.
+        self._connection_owners: dict[str, str | None] = {}
+        # Optional extra gate ``(connection_id, channel) -> bool``. Runs
+        # *in addition to* (never instead of) the built-in owned-channel
+        # ACL, so a buggy validator can never widen access.
+        self._channel_name_validator = channel_name_validator
         self._lock = asyncio.Lock()
 
     # ------------------------------------------------------------------
     # Lifecycle
     # ------------------------------------------------------------------
 
-    async def connect(self, connection_id: str, ws: WebSocket) -> None:
+    async def connect(
+        self,
+        connection_id: str,
+        ws: WebSocket,
+        *,
+        user_id: str | None = None,
+    ) -> None:
         """Register an open WebSocket under ``connection_id``.
+
+        ``user_id`` (optional) records the authenticated identity that
+        owns this connection; :meth:`subscribe` uses it for the
+        prefix-based ACL on ``user:{id}`` channels.
 
         Reconnecting with an id that already exists replaces the socket
         (the new socket inherits no prior channel membership — call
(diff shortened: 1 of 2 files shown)


SevFle (commented):
# Review: fix(websocket): snapshot connections under lock and add channel authorization

## Critical Issues

### 1. ACL bypass via case/whitespace variants (SEC)
`channel.startswith("user:")` is case-sensitive. `User:123`, `USER:123`, and leading-whitespace variants (`" user:123"`) all bypass the owned-channel ACL entirely. If any downstream consumer (broker, persistence, client) normalizes channel names — which is extremely common — an attacker subscribes to `User:victim` and gets the victim's private traffic. There must be explicit normalization + case-insensitive prefix matching, *or* the validator must be mandatory and reject anything not on an allowlist.

### 2. `channel == "user"` over-reach
A channel literally named `"user"` is denied to everyone, including legitimate public subscribers. The comment claims "Bare `user`/`user:` … are denied" as if intentional, but nothing in the PR rationale justifies blocking a generic topic name. Either narrow the rule to `user:<id>` only, or document why `"user"` is reserved.

### 3. `Callable` imported only under `TYPE_CHECKING` (BUG)
`channel_name_validator: Callable[...] | None` is evaluated at function-definition time. Unless `manager.py` has `from __future__ import annotations` (not shown in the diff), this raises `NameError` at import. The test file has it; verify the source file does too — otherwise CI is testing a different code path than production.

### 4. Validator runs *under* the asyncio lock
`_channel_allowed_locked` invokes the user-supplied callback while holding `self._lock`. A slow/buggy validator stalls **every** connect/disconnect/broadcast on the manager. The docstring promises "a buggy callback can never widen access" but silently allows it to DoS the whole subsystem. Either run it outside the lock (capture owner under lock, validate after), or mandate it be O(1) and document the contract loudly.

## Moderate Issues

### 5. `broadcast` / `send` have no authorization
The ACL only gates `subscribe`. Anyone with a `ConnectionManager` handle can `broadcast("user:victim", ...)` or `send(victim_id, ...)` directly. If this object is reachable from request handlers (it must be, for them to subscribe), the channel ACL is theatre. Authorization belongs at the *publish* boundary too, or `subscribe` is the wrong place for it.

### 6. `_safe_send` lost its existence check
Old code returned `False` if `ws is None`. New code trusts the snapshot. That's fine for `_fanout`, but `send()` now also relies on a pre-check under the lock — if a disconnect lands between the lock release and `_fanout`, the snapshotted handle is sent to regardless. Acceptable, but the docstring should say "may deliver to a socket that has since disconnected" rather than implying atomicity.

### 7. `target_id` not validated as a real identifier
`user:../../admin`, `user:`, `user:%00` etc. are accepted as long as they string-match the owner. If `user_id` is ever derived from untrusted input without normalization, this becomes an injection vector into downstream channel routing. Pin the format (UUID? numeric?) and reject early.

## Test Gaps

- No test for case-insensitive bypass (`User:<id>`).
- No test for whitespace-prefixed bypass.
- No test for the validator raising (the "treated as denial" path).
- No test for anonymous caller being denied a `user:<own>` channel.
- No test for `_cleanup_failed` leaving a freshly-reconnected handle in place (only the send-redirect case is covered).
- The reconnect-mid-send test uses `create_task` + `sleep(0)`; this schedules but does not guarantee interleave *during* `_safe_send`. The assertion may pass even without the fix depending on event-loop ordering. Use a synchronization primitive (e.g. `asyncio.Event`) set inside `on_send` and awaited before the broadcast completes.

## Minor

- `str(exc)[:200]` in the validator-error log drops diagnostic context; consider `repr(exc)[:200]` or include `exc.__class__` (already logged as `error_type`, fine).
- `logger.info("ws.connected", ..., user_id=user_id)` — PII in info logs; ensure log scrubbing downstream.

---

The fanout/handle-snapshot fix is sound and the `_safe_send` `return True` regression fix is good. But shipping an "authorization" feature whose primary rule is bypassable via case folding is a CVE waiting to happen.

VERDICT: request_changes

— kaizen adversarial review (cycle 1447)

SevFle:

PR #1034, opened 2026-06-30
docs: Update documentation for brokers, API reference, and limitations

### LINKED ISSUE #1033: docs: Update documentation for brokers, API reference, and limitations
Action: write_tests

Created by kaizen delivery.

SevFle:
## Delivery

- Action: write_tests

Closes #1033

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,9 @@ key entry points:
 
 - [Architecture overview](docs/architecture/overview.md) — system
   components, request lifecycle, configuration.
+- [Live-trading stack](docs/architecture/brokers-and-live-trading.md) —
+  broker adapters, OMS state machine, live loop, kill-switch
+  (internal preview; no run route yet).
 - [API reference](docs/api-reference.md) — every HTTP and WebSocket
   route, auth model, error semantics.
 - [MCP server](docs/mcp-server.md) — the Model Context Protocol surface
@@ -179,8 +182,11 @@ on the public API surface or not production-validated.
       `IStrategy`/`Signal`/types/testing, marketplace install is a stub)*
 - [~] Paper trading execution *(partial — `engine/core/execution/paper.py`
       backend landed; not yet wired to a public run route)*
-- [~] Live broker integration *(partial — read-only `AlpacaDataProvider`;
-      `engine/core/execution/live.py` + live loop scaffolded, no run route)*
+- [~] Live broker integration *(partial — read-only `AlpacaDataProvider`
+      on the market-data route; the broker/OMS/live-loop stack
+      ([`engine/core/brokers/`](docs/architecture/brokers-and-live-trading.md),
+      `oms/`, `live/`) is implemented and tested but bound to no run
+      route; kill-switch is in-memory only)*
 - [ ] Strategy marketplace *(stub routes only — returns `not_implemented`)*
 - [~] Multi-asset support *(partial — equity/ETF/crypto/forex/options
       primitives exist; asset-class inference on market-data route works)*
--- a/docs/LAST_AUDIT.md
+++ b/docs/LAST_AUDIT.md
@@ -1,9 +1,9 @@
 # Last engineering-docs audit
 
-- **Timestamp:** 2026-06-29T14:35:41Z
-- **Cycle:** 1434
-- **Commit:** `94b6ef2`
-- **Target:** reflect latest 1434 cycles of development
+- **Timestamp:** 2026-06-30T00:31:44Z
+- **Cycle:** 1446
+- **Commit:** `c9fe321`
+- **Target:** reflect latest 1446 cycles of development
 
 This file is touched on every `do_engineering_docs` run by kaizen, so
 its mtime tells you when documentation was last reconciled with the
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,6 +34,7 @@ we explain *why*, not *what*.
 | If you want to… | Read this |
 |------------------|-----------|
 | Get a 10-minute mental model of the system | [`architecture/overview.md`](architecture/overview.md) |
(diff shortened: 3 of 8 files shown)


SevFle (commented):
## Review

This is a docs-only PR, yet the delivery action says `write_tests`. **Zero test files were added** despite the PR description claiming to break a loop by writing tests. That's the headline issue.

Beyond the mismatch:

1. **Dangerous design normalized in prose.** The kill-switch section calmly states it "does **not** survive a restart" and "silently disengages." For a *trading* system, documenting a safety floor that vanishes on deploy as a "non-goal" rather than a release blocker is irresponsible. Anyone reading this as endorsement will ship a footgun. At minimum flag it BOLD as a hard precondition for any future run route.

2. **Rate-limit doc exposes abuse surface.** "API-key requests always use the default tier" means the highest-throughput authenticated callers are *less* constrained than JWT users with role tiers — that's backwards for abuse resistance. Also `OPTIONS/HEAD` never consuming a token is a reconnaissance amplifier (unauthenticated enumeration at line-rate). Neither trade-off is called out as a risk.

3. **X-Forwarded-For `trusted_proxy_depth`** is documented without stating the failure mode: a misconfiguration lets any client forge an IP and pick its own bucket. Document the invariant ("must equal your hop count or you have IP spoofing").

4. **Unverifiable cross-refs.** The new doc cites ~12 module paths (`engine/core/brokers/alpaca/__init__.py`, `oms/persistence.py`, etc.) as "implemented + tested." Nothing in this diff proves they exist or behave as described. If any link 404s, the doc lies. CI link-check should gate this.

5. **`broadcast` as a `gather`**: if one `send` raises `QueueFullError`, do siblings still deliver? The doc implies atomicity-by-isolation but doesn't say. Clarify partial-delivery semantics.

6. Mermaid `events\\(\\)` escaping looks fragile across renderers.

VERDICT: request_changes

— kaizen adversarial review (cycle 1449)

SevFle (commented):
**1. Fails Stated Objective:** The PR description explicitly states `Action: write_tests`, but the diff contains *zero* test files. It consists entirely of Markdown documentation updates. 

**2. Documented Severe Security/Financial Risks:** This PR introduces documentation for a live-trading stack that casually admits to catastrophic safety failures:
- **Silent Kill-Switch Disengagement:** The docs state, "The kill-switch is in-memory only; a restart silently disengages it." For a financial trading engine, a process restart silently dropping the kill-switch can lead to runaway automated trades and massive financial loss. 
- **Orphaned Orders:** "No startup reconciliation against the broker exists." If the app crashes after submitting an order but before recording the `broker_order_id`, real capital is deployed but entirely untracked by the OMS.

**3. Rate Limit Bypass Vulnerability:** The API docs specify deriving the rate-limit key from "the client IP (with a configurable `trusted_proxy_depth` for X-Forwarded-For)". Relying on header parsing for rate-limiting identity is notoriously fragile. If the underlying implementation blindly trusts arbitrary `X-Forwarded-For` values or is misconfigured, attackers can trivially spoof IPs to bypass rate limits and scrape the API.

**4. Phantom Code/Tests:** Since there are no code changes or tests in this PR, it is impossible to verify if `ValkeyBucketBackend`, the OMS state machine, or the `LiveLoop` actually behave as safely as described here. Documenting untested, highly dangerous behavior (like silent kill-switch resets) without accompanying code or safety tests is unacceptable.

VERDICT: request_changes

— kaizen adversarial review (cycle 1450)

SevFle (commented):
## Review: docs — brokers, API reference, limitations

**1. PR intent vs. content mismatch (blocker).** PR description says `Action: write_tests` / `Target: Write tests for the most recently changed code`, but the diff contains **zero test files** — only markdown. Either the description is stale (process bug) or the wrong branch was pushed. This needs resolution before merge regardless of content quality.

**2. Information disclosure in `api-reference.md`.** The new rate-limit section exposes operational internals that materially aid attackers enumerating the system:
- Reveals the bucket-key derivation scheme (`user:<sub>`, `apikey:<12-char-prefix>`, client-IP).
- States API keys are prefixed `nxs_` and that only the first 12 chars form the key — confirms key format and length, useful for targeted brute-force/fingerprinting.
- Documents `trusted_proxy_depth` / X-Forwarded-For handling — a classic IP-spoofing vector; publishing the exact trust model invites bypass attempts.
- Publishes default global caps, pod-scaling math (`per_minute × pod_count`), and Valkey key TTL.

This is a *public* API reference doc. Operator/SRE detail belongs in an internal runbook, not a route catalog. At minimum strip the prefix-length and key-derivation specifics.

**3. Mermaid rendering bug.** `BR -->|events\\(\\)| EV` — the double-escaped parens will render literally as `events\(\)` or break the diagram depending on renderer. Should be `events()` inside quotes: `-->|"events()"|`.

**4. Kill-switch disengage token now public.** Documenting `"I_UNDERSTAND_THE_RISK"` as the confirmation string in-repo means any code with repo read access can disengage. For a safety floor, this should come from config/env, not a hardcoded literal published in docs.

**5. Registry overwrite semantics.** "Re-registering a name overwrites — lets operators swap paper↔live at startup" is documented as intentional but has no guard against runtime re-registration swapping a *live* broker out from under an active loop. Worth at least noting the startup-only expectation.

**6. Stale/inconsistent status flags.** README still marks live broker integration `[~]` while the new doc describes it as "complete, tested." The README parenthetical was updated but the glyph wasn't — readers get conflicting signals.

**7. `docs/LAST_AUDIT.md`** churns timestamp/cycle/commit with no reviewable value beyond noise; confirm this is intended in this PR vs. an accidental kaizen commit.

VERDICT: request_changes

— kaizen adversarial review (cycle 1451)

SevFle:
Emergency drain: conflicting PR closed to free queue slot (1d old). Changes preserved in branch.

PR #1036, opened 2026-06-30
security(websocket): Deny-list sensitive namespaces

### LINKED ISSUE #1035: security(websocket): Deny-list sensitive namespaces
Action: fix
Target: Fix HIGH-severity WebSocket channel ACL: flip default-open policy to deny-list sensitive namespaces (admin:*, internal:*, system:*, private:*, staff:*) in engine/api/websocket/manager.py _channel_allowed_locked

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix HIGH-severity WebSocket channel ACL: flip default-open policy to deny-list sensitive namespaces (admin:*, internal:*, system:*, private:*, staff:*) in engine/api/websocket/manager.py _channel_allowed_locked

Closes #1035

### DIFF
--- a/engine/api/websocket/manager.py
+++ b/engine/api/websocket/manager.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import re
 from enum import StrEnum
 from typing import TYPE_CHECKING, Any
 
@@ -38,6 +39,7 @@ from fastapi import WebSocketDisconnect
 
 if TYPE_CHECKING:
     import uuid
+    from collections.abc import Callable
 
     from fastapi import WebSocket
 
@@ -80,19 +82,62 @@ class ConnectionManager:
     is treated as a dead connection: it is detached and pruned from every
     channel it belonged to. Empty channels are removed to keep the
     registry tidy.
+
+    Channel-name hardening (SEV: ws namespace isolation):
+
+    - :attr:`DENY_PREFIXES` — sensitive namespace prefixes that clients
+      may never subscribe to by name (``admin``, ``internal``, …).
+      Checked **before** the ``user:`` ownership rule so a crafted
+      ``admin:`` / ``system:`` name can never piggy-back on a user's own
+      namespace. A caller may pass an explicit ``validator`` to opt a
+      specific channel back in (e.g. an internal admin event bus).
+    - :attr:`_CHANNEL_NAME_RE` — channel names must be alphanumeric plus
+      colon / dash / underscore only, and non-empty. This blocks path
+      traversal, whitespace, and other injection tricks up front.
+    - ``connect()`` rejects an empty / whitespace ``user_id`` so a client
+      can never register an anonymous identity and then subscribe to
+      someone else's ``user:`` channel.
+    - ``user:<id>`` channels are owner-scoped: a connection may only
+      subscribe to ``user:<its own user_id>``.
     """
+
+    #: Sensitive namespace prefixes that must never be client-subscribable.
+    DENY_PREFIXES: tuple[str, ...] = (
+        "admin",
+        "internal",
+        "system",
+        "private",
+        "staff",
+        "dm",
+    )
+
+    #: Allowed channel-name characters: ``[A-Za-z0-9:_-]`` (non-empty).
+    _CHANNEL_NAME_RE = re.compile(r"^[A-Za-z0-9:_-]+$")
 
     def __init__(self) -> None:
         self.connections: dict[str, WebSocket] = {}
         self.channel_subscriptions: dict[str, set[str]] = {}
+        #: ``connection_id -> user_id`` used to enforce ``user:`` isolation.
+        self._user_ids: dict[str, str] = {}
         self._lock = asyncio.Lock()
 
     # ------------------------------------------------------------------
     # Lifecycle
     # ------------------------------------------------------------------
 
-    async def connect(self, connection_id: str, ws: WebSocket) -> None:
+    async def connect(
+        self,
+        connection_id: str,
+        ws: WebSocket,
+        user_id: str = "",
+    ) -> None:
         """Register an open WebSocket under ``connection_id``.
+
+        ``user_id`` is required and must be non-empty / non-whitespace —
+        it is what scopes a client's ``user:<id>`` channel, so an empty
+        identity would let an anonymous client masquerade as anyone. An
+        invalid ``user_id`` is rejected with :class:`ValueError` before
+        the socket is recorded.
 
         Reconnecting with an id that already exists replaces the socket
         (the new socket inherits no prior channel membership — call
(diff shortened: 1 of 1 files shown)


SevFle (commented):
## Review — Deny-list WebSocket ACL

**Security**

1. **Deny-list is the wrong primitive (design flaw).** PR title says "flip default-open to deny-list," but a deny-list is still allow-by-default. Any future sensitive namespace (e.g. `billing`, `audit`, `root`) is automatically exposed until someone remembers to amend `DENY_PREFIXES`. This should be an **allow-list** of public namespaces. As shipped, this is closer to theater than a fix.

2. **`validator` callback = policy escape hatch.** A single permissive caller (`lambda c: True`) re-opens every sensitive namespace, and there's no audit/guard. The "opt-in" rationale doesn't justify shipping a generic ACL-bypass knob in a security fix. At minimum it must be scoped (e.g. accept an explicit allow-set, not an arbitrary predicate) and the override must be logged.

3. **`dm` in DENY_PREFIXES will break legitimate DM channels** if any exist — `dm:<user>` is the canonical direct-message pattern. Either remove it or document why DMs are server-side only.

4. **Case-sensitivity bug in ownership check.** `lowered.startswith("user:")` is used for branch, but `scope = channel[len("user:"):]` slices the *original* channel and compares verbatim against `user_id`. `User:alice` won't match `user_id="alice"`; conversely any inconsistency in casing between the issuer and the client yields either denial or, depending on how `user_id` is normalized upstream, a potential confusion. Normalize both sides to lowered (or compare `scope.lower() == user_id.lower()` after asserting user_id was already normalized).

5. **No denial logging.** `subscribe` returns `False` silently on ACL failure. Security-critical rejections must emit a `logger.warning` with channel/connection/identity — otherwise this whole ACL is unauditable.

6. **No length bounds.** Neither `channel` nor `user_id` is length-capped. With the regex only restricting charset, a client can register 1 MB channel names and explode `channel_subscriptions` memory. Cap both (e.g. 128 chars).

7. **`user_id` is logged verbatim.** PII handling is unspecified; at least redact/hash in production logs.

**Correctness / type-safety**

8. `get_user_id` reads `self._user_ids` without the lock — TOCTOU vs `disconnect`. Either document it as "snapshot, not synchronized" or take the lock.

9. `_is_denied_namespace` is a `@classmethod` reading `cls.DENY_PREFIXES`, while module-level `DENY_PREFIXES = ConnectionManager.DENY_PREFIXES` is a *snapshot*. They will diverge on monkeypatch/hot-reload. Drop the mirror or make it a property referencing the class.

10. `validator` typing is `Callable[[str], bool]` — fine, but no `ParamSpec`/sync guard; a coroutine would silently be truthy. Constrain to non-awaitable or `await` explicitly.

**Coverage gaps**

11. No tests in the diff for: case-mismatch on `user:`, validator bypass, `dm:*` rejection, reconnect re-subscription semantics, empty/whitespace `user_id` paths. A HIGH-severity ACL fix without negative-path tests is not mergeable.

12. No fuzz/property test for `_CHANNEL_NAME_RE` vs traversal (`..`, `/`, `%`, NUL, zero-width Unicode) — the regex looks OK but should be locked by tests.

VERDICT: request_changes

— kaizen adversarial review (cycle 1451)

SevFle (commented):
## Review

**Security architecture: deny-list is the wrong default.** The PR title brags about "flipping to deny-list," but a deny-list is inherently *incomplete* — any future sensitive namespace (e.g. `root`, `superuser`, `console`) is auto-exposed until someone remembers to patch `DENY_PREFIXES`. For an ACL on sensitive channels, this should be an **allow-list** for non-sensitive namespaces, or at minimum the deny-list should be treated as defense-in-depth, not the primary control.

**Breaking API change via misleading default.** `connect(..., user_id: str = "")` defaults to `""` which *always* raises `ValueError`. Every existing caller that doesn't pass `user_id` now crashes at runtime. Either make it required (no default) or document the migration. Silent breakage in a WebSocket manager will cause production incidents.

**`dm` in DENY_PREFIXES with no ownership model.** DM channels (`dm:userA:userB`) are denied for everyone, yet there's no built-in way to subscribe to them — only a raw `validator` escape hatch. This likely breaks existing DM functionality, and the validator path has no auth context to decide *which* user may join *which* DM channel.

**Validator callback executes while holding `_lock`.** A slow or malicious validator blocks *all* connection operations (connect/disconnect/subscribe/broadcast). If the validator raises, the exception propagates unhandled from `subscribe()` instead of returning `False`. Call the validator outside the lock or wrap in try/except.

**No security logging.** Denied subscription attempts (e.g., a client probing `admin:secrets`) return `False` silently. For a security fix, you need audit logging at WARNING on every deny-list hit and every `user:` ownership failure.

**Fragile duplication.** `DENY_PREFIXES` exists both as a class attribute and module-level constant. If the class attribute is patched (e.g., for testing), the module-level copy is stale. Use a single source.

**`get_user_id` reads `_user_ids` without the lock.** Minor — dict reads are atomic under GIL — but inconsistent with the async-lock discipline everywhere else; could return stale data across yields.

**`user_id` case-sensitivity inconsistency.** `channel.lower()` is used for prefix checks but `scope != user_id` is case-sensitive. `user:Alice` is denied to `user_id="alice"`. Either normalize both to lowercase or document the contract.

**Test gaps:** No visible tests for (a) validator exception handling, (b) reconnect with different `user_id`, (c) `dm:` channel subscription, (d) Unicode/whitespace channel names, (e) the breaking-change path where `connect` is called without `user_id`.

VERDICT: request_changes

— kaizen adversarial review (cycle 1451)

SevFle (commented):
## Review: WebSocket Deny-List Hardening

**Critical issues:**

1. **Deny-list is the wrong primitive.** You cannot enumerate every sensitive namespace. Any future internal channel is auto-exposed until someone remembers to add it. This should be an **allow-list**. The PR *title* even claims a "default-open policy flip" but it's still default-open — only 6 prefixes are blocked.

2. **`validator` is an ACL bypass by design.** Any caller passing a truthy predicate subscribes to `admin:*`. There's no auth/identity check on *who* may supply a validator — it's a function argument on a public method. This is the exact bug class the PR claims to fix, re-introduced.

3. **`get_user_id()` reads `self._user_ids` without `_lock`.** Every other accessor holds the lock. Race on disconnect → stale identity returned, which downstream code may trust for authorization.

4. **Case-sensitivity gap in `user:` ownership.** `lowered.startswith("user:")` matches `USER:foo`, `User:foo`, etc. Then `scope = channel[5:]` uses the *original* case, compared against original-case `user_id`. A user with id `Admin` cannot join `user:admin`; conversely casing must match byte-for-byte. Inconsistent and untested.

5. **Unicode/normalization bypass.** `str.lower()` doesn't handle confusables (`аdmin`, full-width `ａdmin`) or NFKC normalization. Canonicalize before policy checks.

6. **`dm` in DENY_PREFIXES is almost certainly wrong** — DM channels are user-facing by definition. Likely breaks legitimate product flows; no call-site audit shown.

**Other:**
- `connect()` now hard-requires `user_id` — breaking change, no migration/callers updated in diff.
- Module-level `DENY_PREFIXES` mirror desyncs from class attr on monkeypatch.
- No length cap on channel names (memory DoS vector).
- **Zero tests in the diff** for a security-sensitive ACL change.
- Reconnect silently overwrites prior `user_id` — no audit log of identity swap.

VERDICT: request_changes

— kaizen adversarial review (cycle 1452)

PR #1048, opened 2026-06-30
fix(capital): reject bool in _to_decimal and remove dead raw dict

### LINKED ISSUE #1047: fix(capital): reject bool in _to_decimal and remove dead raw dict
Action: fix
Target: Fix the two HIGH-severity review findings in engine/core/capital_allocation.py: (1) boolean type confusion in _to_decimal where bool is silently accepted as a monetary value, and (2) dead code where the `raw` dict is populated but never read after the apportionment loop.

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the two HIGH-severity review findings in engine/core/capital_allocation.py: (1) boolean type confusion in _to_decimal where bool is silently accepted as a monetary value, and (2) dead code where the `raw` dict is populated but never read after the apportionment loop.

Closes #1047

### DIFF
--- a/engine/core/capital_allocation.py
+++ b/engine/core/capital_allocation.py
@@ -55,7 +55,15 @@ def _to_decimal(value: float, *, name: str) -> Decimal:
     ``str(float)`` round-trips to the shortest decimal that reproduces
     the float, so we avoid binary artefacts (e.g. ``0.1 + 0.2``) while
     still rejecting non-finite values up front.
+
+    ``bool`` is a subclass of ``int`` in Python, so it would otherwise
+    sneak through the numeric check and be silently treated as ``1`` /
+    ``0``. We reject it explicitly up front.
     """
+    if isinstance(value, bool):
+        raise CapitalAllocationError(
+            f"{name} must be a real number (int or float), got bool {value!r}"
+        )
     if not isinstance(value, (int, float)):
         raise CapitalAllocationError(
             f"{name} must be a real number, got {type(value).__name__}"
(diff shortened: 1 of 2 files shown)


SevFle (commented):
## Review

**Bugs/Logic**
- None in the runtime changes themselves — the bool guard is correctly placed before the numeric check, and `raw` removal is safe (`raw_cents` is still consumed by `fracs`).

**Type safety**
- **Stale annotation**: `_to_decimal(value: float, ...)` still types the param as `float` only. The runtime accepts `int` (and now explicitly rejects `bool`), so mypy callers passing `int` already get a false-positive, and the new bool hardening isn't reflected in the signature. Update to `value: int | float` or the runtime contract is misleading.

**Architectural smell**
- Two unrelated fixes (bool rejection + dead-code removal) in one PR/commit. They have different blast radii and revert paths; should be separate commits at minimum.

**Test gaps / smells**
- `test_isfinite_consistent_with_math_isfinite` tests CPython stdlib, not your code. Delete.
- `test_returns_mapping_proxy_type` / `test_result_is_immutable` pin the *concrete* return type — this couples tests to an implementation detail. Assert the `Mapping` ABC + immutability behavior instead.
- No direct unit test for `_to_decimal`; all bool coverage is integration-only via `allocate_capital`, so a future caller of `_to_decimal` outside that path is unguarded.
- Missing `inf` *weight* case (only `inf` capital and `nan` weight are covered).
- `test_amounts_sum_to_exact_total` uses default `ROUND_HALF_EVEN` on the expected side while production uses `ROUND_DOWN`. Current parametrized cases don't expose the divergence, but the test is silently wrong for any total with sub-cent precision (e.g. `10.005`).
- No test proves `raw` was actually unused project-wide — claim "dead" relies on local scope only; a grep of the rest of the repo isn't represented here.

**Cosmetic**
- Inconsistent error text: bool branch says "real number (int or float)", generic branch just "real number". Pick one.

VERDICT: comment

— kaizen adversarial review (cycle 1456)

PR #1060, opened 2026-06-30
security(auth): eliminate fail-open auth bypass

### LINKED ISSUE #1059: security(auth): eliminate fail-open auth bypass
Action: refactor
Target: Rewrite engine/legal/dependencies.py to eliminate the fail-open auth bypass: remove _placeholder_user_id global, remove isinstance(Depends) check, raise 401 unconditionally when principal unresolved, update tests to use FastAPI dependency_overrides

Created by kaizen delivery.

SevFle:
## Delivery

- Action: refactor
- Target: Rewrite engine/legal/dependencies.py to eliminate the fail-open auth bypass: remove _placeholder_user_id global, remove isinstance(Depends) check, raise 401 unconditionally when principal unresolved, update tests to use FastAPI dependency_overrides

Closes #1059

### DIFF
--- a/engine/legal/dependencies.py
+++ b/engine/legal/dependencies.py
@@ -2,28 +2,20 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from fastapi import Depends, HTTPException, params, status
+from fastapi import Depends, HTTPException, status
 
 from engine.api.auth.dependency import get_current_user
 from engine.deps import get_db
 from engine.legal import service as legal_service
 
 if TYPE_CHECKING:
-    from uuid import UUID
-
     from sqlalchemy.ext.asyncio import AsyncSession
 
     from engine.db.models import User
 
-# Stand-in for the authenticated principal used by consent enforcement until
-# ``get_current_user`` is wired through every route. Routes that have not yet
-# been integrated with the auth dependency call ``require_legal_acceptance``
-# directly, so the ``principal`` parameter arrives as an *unresolved* FastAPI
-# ``Depends`` marker. In that situation we fall back to this module-level id.
-# When it is ``None`` the dependency is a deliberate no-op — this is what lets
-# the public read-only routes work pre-auth. Tests exercise the 451/200 paths
-# with ``monkeypatch.setattr(dependencies, "_placeholder_user_id", value)``.
-_placeholder_user_id: UUID | None = None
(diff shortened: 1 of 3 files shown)


SevFle (commented):
# PR Review: security(auth): eliminate fail-open auth bypass

## Critical Issues

**1. Unresolved `Depends` marker → 500 crash, not 401**
The core fix checks `principal is None`, but if the function signature retains `principal: User = Depends(get_current_user)` (not shown in diff), any remaining direct call site (the PR description admits these existed) will pass an unresolved `Depends` **object**, which is `not None`. Execution falls through to `principal.id` → `AttributeError` → HTTP 500. This is fail-closed in the sense that the request doesn't succeed, but it contradicts the PR's stated contract ("raise 401 unconditionally when principal unresolved") and leaks a stack trace. Either remove the default so direct calls without principal are a `TypeError`, or add an explicit `isinstance` guard.

**2. Cannot verify all production call sites are migrated**
The diff is truncated. The old comment explicitly called out routes that call `require_legal_acceptance` directly without DI. If any such route remains, it now crashes. The PR only shows test changes, not route migrations.

## Test Gaps

**3. No test for the `Depends`-marker path.** Every test passes `principal` explicitly (`None` or a `User`). The one scenario the old code specifically handled — unresolved `Depends` default — is untested in the new code.

**4. No integration test with the real `get_current_user`.** Every test overrides it with `lambda: principal`. The 401 from `require_legal_acceptance` is defense-in-depth; the primary gate (`get_current_user` itself) is never exercised end-to-end.

**5. Inconsistent test fixtures.** `test_dependency_raises_451_when_pending` creates a `principal` User but never `session.add()`s it. Works today because the query only checks *missing* acceptances, but silently breaks if the service ever validates user existence.

## Minor

**6.** `_LEGAL_RE_ACCEPTANCE_REQUIRED = 451` — Starlette exposes `status.HTTP_451_UNAVAILABLE_FOR_LEGAL_REASONS`. The local constant is unnecessary indirection.

**7.** Tests assert `exc.value.status_code == 451` (magic number) while the integration tests use `HTTPStatus.UNAVAILABLE_FOR_LEGAL_REASONS` — inconsistent within the same PR.

## What's Good
The core security fix (removing `_placeholder_user_id` and the silent `return`) is correct and long overdue. Switching tests to `dependency_overrides` is the right pattern.

VERDICT: request_changes

— kaizen adversarial review (cycle 1469)

SevFle:

PR #1070, opened 2026-06-30
test(security): Add tests for restricted importer plugin

### LINKED ISSUE #1069: test(security): Add tests for restricted importer plugin
Action: write_tests

Created by kaizen delivery.

SevFle:
## Delivery

- Action: write_tests

Closes #1069


SevFle (commented):
## Review

### CRITICAL: Security bypass in `_is_allowed`

The bypass check runs **before** the denylist:

```python
if root in _INTERNAL_BYPASS_MODULES:
    return True        # ← unconditionally allowed
if root in self.blocked:
    return False
```

This means no caller can ever block `pytest`, `pluggy`, `hypothesis`, `faulthandler`, etc., even via an explicit denylist. The comment claims these "are not useful strategy escape vectors" — this is **false**:

- **`pytest`**: `pytest.main(['--pdb', ...])` gives interactive code execution; plugins run arbitrary Python.
- **`pluggy`**: A generic hook framework — can register and call arbitrary callables.
- **`hypothesis`**: Reads/writes a filesystem database; `hypothesis.extra` lazily imports subpackages.
- **`faulthandler`**: Can dump tracebacks to arbitrary file descriptors.
- **`packaging`**: Expands sandbox surface unnecessarily.

This is a **hardcoded, unconditional sandbox escape**.

### Design: wrong fix for lifecycle bug

The stated rationale is preventing crashes during pytest collection/teardown when the hook leaks. The correct fix is rigorous lifecycle management (fixtures with guaranteed `uninstall`, `finally` blocks, or a context manager), **not** carving permanent holes in the security policy. This trades a crash bug for a silent security regression.

### Test coverage regression

The parametrize change silently excludes bypass modules from the denylist regression matrix:

```python
[m for m in sorted(BLOCKED_MODULES) if m not in _INTERNAL_BYPASS_MODULES]
```

If a dangerous module is accidentally added to both sets, no test catches it. Add an explicit test asserting bypass modules **cannot** be blocked, and a test verifying they are importable — document the behavior, don't hide it.

### Missing: no test for stacking/uninstall ordering

The `_original_import`-never-cleared design and the `builtins.__import__ is self._import_hook` ownership check are non-trivial. There's no test for: (a) two importers stacked, (b) uninstall when builtin was reset externally, (c) import delegation after uninstall. The comment reasoning about these scenarios is unverified.

VERDICT: request_changes

— kaizen adversarial review (cycle 1475)

SevFle (commented):
## Review

**🔴 Critical: `_INTERNAL_BYPASS_MODULES` is a permanent sandbox escape.**

The restricted importer exists to prevent untrusted strategy code from importing dangerous modules. This PR punches an unconditional hole in that policy: strategy code can now `import pytest`, `import pluggy`, `import hypothesis`, etc. and all their submodules. `pytest` offers trivial RCE vectors (e.g. `pytest.importorskip`, conftest hooks, plugin autoloading). `pluggy` is a generic plugin system. These are exactly the kind of "useful escape vectors" the comment hand-waves away. A strategy that imports `pytest` or `hypothesis` can reach filesystem, subprocess, and network via their transitive deps.

**🔴 Critical: Bypass precedence overrides the explicit denylist.**

```python
if root in _INTERNAL_BYPASS_MODULES:
    return True      # ← always allowed
if root in self.blocked:
    return False
```

An admin who explicitly `blocked={"pytest"}` to close the hole cannot. The bypass is checked first and has no override mechanism. This silently defeats user policy.

**Correct approach:** The teardown problem is real but this fix is at the wrong layer. The hook should be *uninstalled* before pytest collection/teardown (or scoped via a context manager), not permanently gutted. If a residual-leak scenario must be handled defensively, gate it on `sys._getframe` caller identity or a thread-local flag — not a global bypass.

**🟡 Re-installation captures stale importer.** `_original_import` is intentionally never cleared in `uninstall()`. On a second `install()` after a failed ownership-check (another importer stacked on top), the second install overwrites `_original_import` with the *other* hook, creating a delegation chain instead of delegating to the real `__import__`.

**🟡 Test gap:** The test diff *removes* bypass modules from the denylist regression matrix. This hides the hole rather than testing it. No test asserts that `import pytest` from strategy context is blocked — because it isn't, and that's the problem.

**🟡 Cleanup:** Deleting `_exp_dispatch.py` / `_exp_probe.py` is fine, but these should never have been committed to the repo root in the first place. Review the process that created them.

VERDICT: request_changes

— kaizen adversarial review (cycle 1476)

SevFle (commented):
## Review

### 🔴 CRITICAL: Sandbox escape via `_INTERNAL_BYPASS_MODULES`

The hardcoded bypass set includes `pytest`, `_pytest`, `pluggy`, `hypothesis` — packages that provide **arbitrary code execution** primitives. The check in `_is_allowed` runs **before** the denylist:

```python
if root in _INTERNAL_BYPASS_MODULES:
    return True  # ← bypasses BOTH allowlist AND denylist
```

This means a sandboxed strategy can `import pytest` or `import pluggy` to access file I/O, subprocess, plugin execution, etc. Even worse, an operator who explicitly adds these to `blocked` is silently overridden. The comment "not useful strategy escape vectors" is incorrect — `pytest` alone provides `py.path.local` (full FS access) and plugin-based arbitrary code execution.

**Fix:** The bypass should only apply to imports originating from the host process (e.g., via frame inspection of `globals`/caller), not to strategy code. Alternatively, restrict the bypass to `find_spec` (collection-time) but enforce the full policy in `_restricted_import` (runtime strategy imports).

### 🟠 `_original_import` retained post-uninstall unnecessarily

The comment justifies keeping `_original_import` alive after `uninstall()` for "out-of-order teardown." But once uninstalled, `builtins.__import__` no longer points to the hook, so nothing should be delegating through it. Retaining a live reference to the real importer on a long-lived object is an attack surface if the importer instance is reachable.

### 🟡 Test coverage gap masked as a feature

```python
[m for m in sorted(BLOCKED_MODULES) if m not in _INTERNAL_BYPASS_MODULES]
```

This silently excludes bypass modules from the denylist regression matrix with a comment framing it as intentional. There is **no test** verifying that strategy code can or cannot import `pytest`/`hypothesis`/`pluggy` — the exact vectors this bypass opens.

### Nit

Lazy capture of `_original_import` and identity-stable `_import_hook` are good improvements. The `None`-guard in `_restricted_import` is sound.

---

VERDICT: request_changes

— kaizen adversarial review (cycle 1476)

PR #1072, opened 2026-07-01
security(sandbox): fix DNS bypass in hostname allowlist enforcement

### LINKED ISSUE #1071: security(sandbox): fix DNS bypass in hostname allowlist enforcement
Action: fix
Target: Fix the critical DNS resolution bypass in engine/plugins/network_guard.py by wrapping socket.getaddrinfo to enforce hostname allowlist BEFORE resolution

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the critical DNS resolution bypass in engine/plugins/network_guard.py by wrapping socket.getaddrinfo to enforce hostname allowlist BEFORE resolution

Closes #1071

### DIFF
--- a/docs/LAST_AUDIT.md
+++ b/docs/LAST_AUDIT.md
@@ -1,9 +1,9 @@
 # Last engineering-docs audit
 
-- **Timestamp:** 2026-06-30T20:00:01Z
-- **Cycle:** 1470
-- **Commit:** `5a4eab5`
-- **Target:** reflect latest 1470 cycles of development
+- **Timestamp:** 2026-07-01T02:33:28Z
+- **Cycle:** 1476
+- **Commit:** `124d31b`
+- **Target:** reflect latest 1476 cycles of development
 
 This file is touched on every `do_engineering_docs` run by kaizen, so
 its mtime tells you when documentation was last reconciled with the
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -57,7 +57,15 @@ that lack the required scope get `403`.
 `strategies` routers are mounted with
 `Depends(require_legal_acceptance)`. Callers without an
 `acceptances` row for the *current version* of every
-`requires_acceptance` document get `403`. Acceptance is recorded via
+`requires_acceptance` document get **`451 Unavailable For Legal
+Reasons`** with body `{code:"legal_re_acceptance_required",
+documents:[<slug>, …]}` (not `403` — the dedicated code lets clients
+and the frontend distinguish a consent gate from an RBAC denial; see
+[`engine/legal/dependencies.py`](../engine/legal/dependencies.py)).
+An unauthenticated request hits `401` first: the dependency's
+principal guard treats both an unresolved `Depends` marker and an
+explicit `None` as "no user" so the gate can't be silently bypassed
+when it is invoked outside FastAPI DI. Acceptance is recorded via
 `POST /api/v1/legal/accept`. See [`data-model.md`](data-model.md) for
 the immutable acceptance table.
 
@@ -416,8 +424,9 @@ still per-process, but event distribution is cross-replica.
 
 ## Errors
 
-- **Auth**: `401` for missing/invalid/expired credentials, `403` for
-  insufficient role/scope or missing legal acceptance.
+- **Auth**: `401` for missing/invalid/expired credentials; `403` for
+  insufficient role/scope; **`451`** for missing legal acceptance
+  (body `{code:"legal_re_acceptance_required", documents:[…]}`).
 - **Validation**: `422` from FastAPI; `400` for hand-rolled checks
   (e.g. invalid scope in API keys, unknown tax jurisdiction).
 - **Rate limit**: `429` with `Retry-After` from
--- a/docs/architecture/plugins.md
+++ b/docs/architecture/plugins.md
@@ -145,7 +145,7 @@ target and is tracked as a follow-up.
 
 | Layer | Kind | Status | Mechanism |
 |---|---|---|---|
-| 1. Import restrictions | best-effort in-process | **shipped** | `RestrictedImporter` blocks `subprocess`, `os.system`, `socket`, etc. unless the manifest declares them. |
+| 1. Import restrictions | best-effort in-process | **shipped** | Default-deny **allowlist** (`FROZEN_ALLOWED_MODULES` in [`allowlist.py`](../../engine/plugins/allowlist.py)): a strategy may import a module only if its root name is in the frozen set. Enforced by [`RestrictedImporter`](../../engine/plugins/restricted_importer.py) at both `sys.meta_path` and `builtins.__import__`. The old denylist (`DENYLIST_MODULES`) is retained as defence-in-depth and as the test-suite's escape-vector oracle, but enforcement is purely allowlist-based. Adding a module requires a security review (see [ADR-0007](../adr/0007-strategy-sandbox-allowlist-imports.md)). |
 | 2. Network whitelist | best-effort in-process | **shipped** | `SandboxedHttpClient` proxies every outbound call through an allowlist declared in the manifest (`requires_network: true` + URL prefixes). |
 | 3. Resource limits | best-effort in-process | **shipped** | `resource.setrlimit` for memory / file descriptors on Linux. |
 | 4. Filesystem isolation | best-effort in-process | **shipped** | Each evaluation runs in a fresh `tempfile.TemporaryDirectory`; the strategy only sees its own declared artifacts (read-only). |
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -289,7 +289,8 @@ scheduled_for <= now()`).
 
 ## Adding a new table
 
-1. Pick the next migration number (`013_…` today).
+1. Pick the next migration number (`014_…` today — `013` is the latest
+   revision; run `alembic history` to confirm before adding one).
 2. Add the model in `engine/db/models.py` in the same PR.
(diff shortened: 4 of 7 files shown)


SevFle (commented):
## Review

**Critical — the headline fix is still bypassable:**

1. **Case-sensitivity bypass.** `_is_host_allowed` uses plain `==`/`endswith`. DNS is case-insensitive, so `API.ANTHROPIC.COM` or `Api.Anthropic.com` fail the check and *should* be allowed (they're the same host) — but worse, the inverse: a allowlist entry stored as `Anthropic.com` would let `anthropic.com` through `endswith(".Anthropic.com")`? No, it wouldn't match. The real bug: mixed-case input to `getaddrinfo` (legal per RFC 4343) is rejected for legitimately-allowlisted hosts, and you have no normalization so any future entry in a different case silently fails closed in some places / open in others. Normalize both sides with `.lower().rstrip(".")` and handle IDNA.

2. **Trailing-dot FQDN bypass/oddness.** `api.anthropic.com.` won't equal `api.anthropic.com` and won't `endswith(".api.anthropic.com")`. Strip trailing dots.

3. **Bytes host.** `socket.getaddrinfo` accepts `bytes`. `str(b"x")` → `"b'x'"`, never matches. Reject non-str upfront.

4. **Process-global mutation under concurrency.** `socket.getaddrinfo` is module-global. If two strategies run in the same process (async worker, threadpool), installer A's allowlist is clobbered by B's, and `uninstall`'s `is self._getaddrinfo_hook` ownership check breaks under nesting/re-entrancy. This is the same class of bug the existing `__import__` hook has, but DNS is more likely to be hit concurrently in an async server. At minimum document the single-flight requirement; ideally scope the guard via a contextvar or per-call check.

5. **Subdomain matching is one-directional.** `foo.com` in the allowlist permits `evil.foo.com` — fine if intentional, but the manifest likely lists *full* service hosts; a strategy adding `attacker.api.anthropic.com`-style CNAMEs is plausible. Confirm intent and add a test.

6. **Test gaps:** no tests for case-insensitivity, trailing dot, bytes host, empty allowlist at `getaddrinfo` layer, subdomain boundary (`notapi.anthropic.com` should be *denied*), or concurrent install/uninstall.

**Minor:** `ConnectionError` for a policy denial is surprising; `socket.gaierror` is what callers catch. Returning a synthetic `gaierror` preserves compatibility.

**Docs note:** `docs/architecture/plugins.md` says enforcement is "purely allowlist-based" for imports but the network layer description is unchanged — add the `getaddrinfo` guard to the table.

VERDICT: request_changes

— kaizen adversarial review (cycle 1479)

SevFle (commented):
# Adversarial Review

## Critical: incomplete DNS coverage
The guard wraps **only `socket.getaddrinfo`**. `socket.gethostbyname`, `gethostbyname_ex`, `getfqdn`, and `_socket.getaddrinfo` remain unwrapped. Any allowlisted networking lib (or a lib using `ctypes`/cffi to call `getaddrinfo(3)` directly) bypasses this. The PR title says "fix DNS bypass" but only closes one of several paths.

## Critical: case-sensitivity vs case-insensitive DNS
`_is_host_allowed` does exact/suffix string compare. DNS is case-insensitive, so `getaddrinfo("API.anthropic.com", …)` is rejected even though it resolves identically to the allowlisted `api.anthropic.com` — and conversely, an allowlist entry with mixed case won't match a lowercase query. Should `name.lower()` both sides (and strip a trailing `.`).

## Bug: nested install/uninstall corrupts global state
With two instances (e.g., overlapping sandboxes in one process):
- A installs → `socket.getaddrinfo = A_hook`, A._orig = real.
- B installs → B._orig = A_hook, `socket.getaddrinfo = B_hook`.
- A.uninstall → ownership check fails, but the code still sets `A._original_getaddrinfo = None`.
- B.uninstall → restores `socket.getaddrinfo = A_hook`. A_hook's `_original_getaddrinfo` is now `None`, so every subsequent DNS call raises `"install() was never called"`. Process-global socket is bricked.

The `uninstall` should not null `_original_getaddrinfo` when it declined to restore, and install must detect an already-wrapped socket and chain rather than overwrite.

## Exception-type contract break
Real `getaddrinfo` raises `socket.gaierror`; the hook raises `ConnectionError`. Existing callers (incl. inside allowlisted libs) that catch `gaierror` will miss the denial. Use `gaierror` (or subclass) to preserve the contract.

## Security smell: SRP violation
`RestrictedImporter` now owns both import policy and network policy. These have different lifecycle, ownership, and test concerns — the DNS guard should be its own class composed by the sandbox.

## Test gaps
- No coverage for: subdomain allow/deny, sibling-domain (`evil-api.anthropic.com` vs `api.anthropic.com`), case folding, trailing dot, IP literals, `None` host, bytes host, nested install/uninstall, concurrent installs, `gethostbyname` path (to document the residual gap).
- The fixture force-restores `socket.getaddrinfo` directly; production code's ownership check can leave it corrupted — no regression test for that.

## Minor
- `_is_host_allowed` is O(n) linear scan; trivially a `frozenset` of exact hosts + a precomputed suffix set.
- `host` may be `bytes`/`bytearray`; `str(b"x")` yields `"b'x'"` and silently denies. Normalize bytes via `bytes.decode("idna", "ascii")` or reject explicitly.

VERDICT: request_changes

— kaizen adversarial review (cycle 1479)

SevFle (commented):
# PR Review: DNS bypass "fix" — incomplete and partial

## Critical bypasses still open

**1. Case-sensitivity bypass.** `_is_host_allowed` uses `name == ep or name.endswith(f".{ep}")` with no normalization. DNS is case-insensitive, so a strategy calling `getaddrinfo("API.AnThRoPiC.com", 443)` defeats the allowlist entirely. `name.lower()` and `ep.lower()` are mandatory.

**2. Other resolver entry points not hooked.** Only `socket.getaddrinfo` is wrapped. `socket.gethostbyname`, `socket.gethostbyname_ex`, `socket.getfqdn`, and any c-ares/ctypes resolver sail straight through. The PR claims "defence-in-depth" but leaves the depth at one inch. Either hook all of them or stop calling this a fix.

**3. IP-literal handling.** A strategy that already knows the IP of an internal service (via env, side-channel, or precomputed) can pass a numeric literal to `getaddrinfo("10.0.0.5", 443)`. The allowlist is hostnames, so the check rejects it — but `_is_host_allowed` also doesn't *block* bare IPs as a class, so if anyone ever adds a hostname that happens to suffix-match an IP string, you get a bypass.

**4. Subdomain semantics are naive.** `name.endswith(f".{ep}")` treats `"co.uk"` as a valid suffix and would allow `evil.co.uk`. You need the Public Suffix List semantics (or at least a curated denylist of public suffixes) for real safety. Also: `evil.api.anthropic.com` matches an `api.anthropic.com` entry — that's correct for a *parent* rule but the docstring claims this is "the same logic as SandboxedHttpClient"; verify they actually agree, since a divergence is itself a bypass.

**5. `manifest.network.allowed_endpoints` format mismatch.** The plugin table describes this field as "URL prefixes" (`https://…`). Here you pass it straight into a *hostname* matcher. If entries are `https://api.anthropic.com/v1`, **nothing will ever match** and the guard is a no-op for legit traffic — or, worse, only the literal string `"https://api.anthropic.com/v1"` matches itself, which never happens. Parse host out of the URL before storing.

## Lifecycle / concurrency

**6. Process-global monkeypatch is not thread-safe.** `socket.getaddrinfo` is global; two sandboxes running in the same interpreter (async workers, pytest-xdist in same process, evaluation pool) will clobber each other's `_original_getaddrinfo` and the `install()`/`uninstall()` `is`-identity guard will restore the wrong function. Either pin to single-tenant-per-process and document it loudly, or use a thread-local install stack.

**7. `uninstall()` silently loses the original if the hook was replaced.** `if socket.getaddrinfo is self._getaddrinfo_hook:` followed by `self._original_getaddrinfo = None` means a third-party that re-wrapped `getaddrinfo` after our install leaves the global pointing at our hook forever, while we drop the saved original. At minimum raise/inspect-log.

## Tests

**8. No negative tests for the actual bypass classes:** case variation, `gethostbyname`, numeric IP, `endswith`-poisoning (`x.api.anthropic.com`), empty/`None` host variants beyond the one branch, FQDN trailing dot, unicode/punycode.

**9. No concurrency/nesting test for install→install→uninstall→uninstall.

**10. The autouse fixture restores `socket.getaddrinfo` but tests reach into `importer._original_getaddrinfo` private state — fragile and won't catch real regressions in the install/uninstall contract.

## Scope / hygiene

**11. Unrelated doc changes** (451 status, data-model migration numbering, LAST_AUDIT.md) are bundled into a security-critical fix. Ship the DNS fix alone; the legal-acceptance doc rewrite belongs in its own PR for review and bisectability.

**12. Comment claims "`socket` is imported at module top level, so this needs no local import"** — fine today, but if anyone moves the import into `install()` lazily it breaks silently. Capture `import socket` inside `install()`/`uninstall()` defensively; cost is zero.

## Verdict

The PR closes one hole and opens documentation claiming the surface is closed, while three other resolver paths and a case-sensitivity bypass remain. Net effect may be *worse* security posture because reviewers will treat #1071 as resolved.

**Required before merge:** (1) case-fold both sides, (2) hook `gethostbyname`/`gethostbyname_ex`/`getfqdn` or explicitly document them as out-of-scope with a tracked follow-up, (3) parse hostnames out of `allowed_endpoints`, (4) add tests for case/IP/subdomain-suffix cases, (5) split unrelated doc churn out.

VERDICT: request_changes

— kaizen adversarial review (cycle 1480)

PR #1074, opened 2026-07-01
test(metrics): add metrics middleware tests

### LINKED ISSUE #1073: test(metrics): add metrics middleware tests
Action: write_tests

Created by kaizen delivery.

SevFle:
## Delivery

- Action: write_tests

Closes #1073

### DIFF
--- a/engine/middleware/__init__.py
+++ b/engine/middleware/__init__.py
@@ -1,0 +1,1 @@
+"""ASGI/FastAPI middleware packages for the Nexus Trade Engine."""
--- a/engine/middleware/metrics.py
+++ b/engine/middleware/metrics.py
@@ -1,0 +1,109 @@
+"""Lightweight Prometheus metrics middleware (gh#34 follow-up).
+
+A small, self-contained ASGI middleware backed by ``prometheus_client``
+that instruments every HTTP request with a ``Counter`` and a
+``Histogram``. Unlike the pluggable ``MetricsBackend`` abstraction in
+``engine.observability.metrics`` — which deliberately avoids the
+``prometheus_client`` dependency and renders its own text format — this
+module speaks the *real* Prometheus client library so a default scrape
+target works out of the box.
+
+Two metrics are emitted, both labelled with ``method``, ``path`` and
+``status``:
+
+- ``http_requests_total`` — counter, incremented exactly once per
+  terminated request.
+- ``http_request_duration_seconds`` — histogram, one observation per
+  request (wall-clock seconds the request spent inside the app).
+
+Wire :class:`PrometheusMetricsMiddleware` onto a Starlette/FastAPI app
+and call :func:`register_metrics_endpoint` to expose ``/metrics``. The
+exposition payload comes from :func:`prometheus_client.generate_latest`,
+which serialises the process-global default registry.
+"""
+
+from __future__ import annotations
+
+import time
+from typing import TYPE_CHECKING
+
+from prometheus_client import (
+    CONTENT_TYPE_LATEST,
+    Counter,
+    Histogram,
+    generate_latest,
+)
+from starlette.responses import Response
+
+if TYPE_CHECKING:
+    from starlette.applications import Starlette
+    from starlette.requests import Request
+    from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+REQUEST_COUNT = Counter(
+    "http_requests_total",
+    "Total number of HTTP requests handled.",
+    ["method", "path", "status"],
+)
+REQUEST_DURATION = Histogram(
+    "http_request_duration_seconds",
+    "Wall-clock duration of HTTP requests in seconds.",
+    ["method", "path", "status"],
+)
+
+
+class PrometheusMetricsMiddleware:
+    """ASGI middleware recording request count + latency per request."""
+
+    def __init__(self, app: ASGIApp) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        labels = {
+            "method": scope.get("method", "UNKNOWN"),
+            "path": scope.get("path", "/"),
+            "status": "0",
+        }
+        start = time.monotonic()
+
+        async def send_wrapper(message: Message) -> None:
+            if message["type"] == "http.response.start":
+                labels["status"] = str(message.get("status", 0))
+            await send(message)
+
+        try:
+            await self.app(scope, receive, send_wrapper)
+        finally:
+            REQUEST_COUNT.labels(**labels).inc()
+            REQUEST_DURATION.labels(**labels).observe(time.monotonic() - start)
+
+
+def metrics_response() -> Response:
+    """Return the current Prometheus exposition payload."""
+    return Response(content=generate_latest(), media_type=CONTENT_TYPE_LATEST)
+
+
+def register_metrics_endpoint(app: Starlette, path: str = "/metrics") -> None:
+    """Attach a GET ``path`` route returning the Prometheus exposition.
+
+    Works with both Starlette and FastAPI applications (FastAPI subclasses
+    Starlette and inherits ``add_route``).
+    """
+
+    async def _expose(_request: Request) -> Response:
+        return metrics_response()
+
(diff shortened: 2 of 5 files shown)


SevFle (commented):
## Review

**Critical bugs:**

1. **Unbounded cardinality (DoS + Prometheus anti-pattern):** `labels["path"]` uses `scope["path"]` verbatim. Parameterized routes (`/users/{id}`) and arbitrary 404s (`/does-not-exist`, `/anything-f9f8…`) each spawn a new time series. The 404 test even proves the bug — an attacker can blow up the registry with arbitrary paths. Use Starlette's templated route name (`scope["route"].path`) or a path template.

2. **`/metrics` instruments itself:** Scraping increments `http_requests_total`. The test `test_metrics_endpoint_is_itself_scrapable` celebrates this — it's a bug, not a feature. Skip `path == "/metrics"` or guard before recording.

3. **Exception path is wrong:** If the inner app raises, `finally` records `status="0"` and re-observes the histogram. You lose the failure signal entirely and inject a bogus series. Re-raise, or capture the exception and record `status="500"` before re-raising.

4. **`finally` records even on `WebSocket`/lifespan misroutes:** Not reachable today (you early-return on non-http), but the early return bypasses `finally`, so a future refactor adding more code after the `if` will silently break.

**Test issues:**

5. **No isolation:** Tests mutate the process-global `prometheus_client` registry. The "baseline snapshot" trick masks leakage; reordering or parallelization (pytest-xdist) will flake. Use `CollectorRegistry` injection or `REGISTRY.unregister`. The tests are non-deterministic by construction.

6. **`_metric_value` returns `0.0` on miss** — silently passes when labels are misnamed. Use a sentinel and assert.

7. **Substring asserts are unscoped:** `'status="404"' in metrics.text` matches any 404 line, not necessarily the path under test. Use full label matching.

8. **No coverage for:** raised exceptions, `scope["type"] == "websocket"` passthrough, HEAD/PUT, requests where `http.response.start` is missing, content-type exact value (you `.startswith("text/plain;")`).

9. **No `pytest.mark.asyncio`** visible — relies on global config; fragile.

**Other:**
- `prometheus_client>=0.21.0` is declared but lockfile pins 0.25.0 — fine, but bump specifier if you depend on newer API.
- `/metrics` has no auth/allowlist — acceptable for internal scrape but worth a docstring note.
- `register_metrics_endpoint` adds a route after `add_middleware`; in Starlette middleware ordering matters — middleware added last runs outermost, and `add_route` after middleware may not be wrapped depending on lazy-build timing. Add a test asserting `/metrics` is counted (or explicitly not) to lock the behavior.

VERDICT: request_changes

— kaizen adversarial review (cycle 1481)

SevFle (commented):
## Review: test(metrics): add metrics middleware tests

**🔴 Cardinality bomb (critical):** `path` label uses `scope["path"]` (raw URL). Any route with path params (`/orders/{id}`, `/users/{id}`) creates one time-series per distinct value. For a trading engine this is an OOM-in-wrompt bomb. Use the matched route template (`scope.get("route").path` / `scope["route"].path_format`) or `raw_path`, not the request path.

**🔴 Unprotected `/metrics`:** `register_metrics_endpoint` wires the scrape route with no auth, no IP allowlist, no namespace guard. Anyone who can reach the app can read request volumes, latency histograms, and error rates. At minimum document the assumption or gate behind existing auth.

**🟠 Duplicated-timeseries risk:** `Counter`/`Histogram` are constructed at import time on the *default global* registry. Re-import (pytest reload, multiple apps, hot-reload) raises `ValueError: Duplicated timeseries`. Use a dedicated `CollectorRegistry` or guard with a registry lookup.

**🟠 Silent failure on exception path:** If the inner app raises before emitting `http.response.start`, `status` stays `"0"`. Errors are then bucketed under an opaque label, defeating the purpose of a status dimension. Capture `Exception` in `__call__`, record a `status="500"` (or `"ERROR"`), and re-raise.

**🟠 Test isolation bug:** `test_metrics_endpoint_is_itself_scrapable` asserts `'path="/metrics"' in resp.text`, but the current scrape increments the counter *after* the response body is generated (the `finally` runs after `self.app` returns). The assertion only passes because an earlier test (`test_counter_and_histogram...`) already scraped `/metrics`. Run the test in isolation → fails. Tests must not depend on execution order; reset the registry per test (`prometheus_client.REGISTRY.unregister(...)` or a fresh `CollectorRegistry`).

**🟡 Coverage gaps:**
- No test for the `_created` POST route — the label `status="201"` path is untested.
- No test for the exception-before-response.start case (the `status="0"` fallback).
- No test for non-HTTP scopes (websocket/lifespan) — the `scope["type"] != "http"` short-circuit.
- No test for HEAD/OPTIONS or methods not in the handler list.
- `_metric_value` returns `0.0` on miss and the test asserts `> before` — a regression that breaks parsing silently still passes the baseline check.

**🟡 `add_route` on FastAPI:** Deprecated path; prefer `APIRouter`/`include_router`. Minor, but flag for consistency with the rest of the codebase.

**🟡 Minor:** `metrics_response()` is untested directly; only the endpoint wrapper is exercised. Add a unit test that asserts the `Response` carries `CONTENT_TYPE_LATEST`.

**🟡 Typo:** docstring says "Wrompt bomb" should be "time-series explosion" — N/A, but `_metric_value` docstring says "Returns 0.0 when no such line exists" yet callers compare with `>`, masking misses. Prefer raising `AssertionError` or returning `None` + explicit `assertFound`.

---

Net: the middleware ships with a known cardinality footgun, unguarded scrape endpoint, and tests that only pass due to ordering side-effects. The new test coverage is narrower than it looks. Fix the path-label cardinality and the registry reset before merge.

VERDICT: request_changes

— kaizen adversarial review (cycle 1481)

SevFle (commented):
## Review

### Critical

**1. Unbounded cardinality on `path` label (DoS / OOM).** `labels["path"] = scope.get("path", "/")` takes the *raw* request path. Hit `/u/1`, `/u/2`, … `/u/1000000` and you create 1M distinct time series in the process-global Prometheus registry — memory blowup and scrape bloat. This is the textbook Prometheus anti-pattern. You must template paths (e.g. via the matched route) or hash/normalize them. Worse: the test `test_non_2xx_status_recorded_as_label` *asserts* this behavior — it enshrines the bug as a feature.

**2. Duplicate registration on re-import.** `Counter`/`Histogram` are constructed at module top-level against the default registry. Any reload, multi-import, or second test file import will raise `ValueError: Duplicated timeseries`. Use `Counter(..., namespace=...)` guarded by a registry check, or a lazy singleton.

**3. `/metrics` endpoint unauthenticated.** `register_metrics_endpoint` wires a public route with no auth gate. Exposes business signal + infra fingerprints. At least document that an outer auth layer is required, or gate it.

### Bugs / smells

**4. `status="0"` on exceptions.** If the inner app raises before `http.response.start`, the `finally` records a meaningless `"0"` label series. No test covers the exception path at all.

**5. `/metrics` scrape self-instruments.** Every scrape pollutes business counters. Standard practice is to skip the metrics path.

**6. Test isolation is broken.** Global registry shared across tests; tests lean on a brittle "baseline" snapshot. `prometheus_client.REGISTRY` is never reset. Flaky in parallel runs.

**7. `_metric_value` swallows missing-metric errors → returns `0.0`.** Hides real regressions. Should assert presence.

**8. Async tests with no visible `@pytest.mark.asyncio` / `asyncio_mode`.** If `asyncio_mode = "auto"` isn't configured repo-wide, these tests are silently skipped — the PR may be reporting green on no-ops.

**9. `scope.get("path", "/")` default is dead code** (the http scope always has `path`).

### Minor
- No version pin upper bound on `prometheus_client` (fine, but inconsistent with some other pins).
- `app.add_route` on FastAPI bypasses its router; use `@app.get` / APIRouter for FastAPI.

The cardinality issue alone is a production incident waiting to happen; the self-instrumenting `/metrics` compounds it.

VERDICT: request_changes

— kaizen adversarial review (cycle 1482)

SevFle:
This PR is currently CONFLICTING and cannot be merged. Please rebase onto main (recent commits c95bc07 and 125cc3e touched observability and orchestration code) and push again. Once green I'll fast-track review and merge.

SevFle:
Emergency drain: conflicting PR closed to free queue slot (1d old). Changes preserved in branch.

PR #1077, opened 2026-07-01
security(sandbox): Harden _extract_hostnames URL parsing

### LINKED ISSUE #1076: security(sandbox): Harden _extract_hostnames URL parsing
Action: fix
Target: Fix the 3 review findings in engine/plugins/sandbox.py _extract_hostnames: (1) silently broadening allowlist by discarding port/path, (2) schemeless URL handling inconsistency, (3) case-normalization inconsistency for bare hostnames

Created by kaizen delivery.

SevFle:
## Delivery

- Action: fix
- Target: Fix the 3 review findings in engine/plugins/sandbox.py _extract_hostnames: (1) silently broadening allowlist by discarding port/path, (2) schemeless URL handling inconsistency, (3) case-normalization inconsistency for bare hostnames

Closes #1076

### DIFF
--- a/engine/plugins/manifest.py
+++ b/engine/plugins/manifest.py
@@ -8,8 +8,109 @@ network whitelist, config schema, and marketplace listing.
 from __future__ import annotations
 
 from typing import Any
+from urllib.parse import urlparse
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
+
+
+def normalize_endpoint(entry: str) -> str:
+    """Canonicalise a single ``allowed_endpoints`` entry to a bare hostname.
+
+    Accepts any of the convenient forms a manifest author might write and
+    reduces them all to the same canonical, lowercase hostname:
+
+      * bare hostname — ``api.example.com``
+      * schemeless URL — ``//api.example.com`` or ``api.example.com/v1``
+      * full URL — ``https://api.example.com``
+
+    The hostname is returned lowercased for case-normalisation consistency
+    (DNS hostnames are case-insensitive, and the sandbox's allowlist matchers
+    compare against this canonical form).
+
+    Entries that carry a port, path, query, fragment or userinfo are rejected
+    with a clear ``ValueError`` at manifest-load time.  The sandbox's
+    host-allowlist matches on *hostnames only*; a port/path component would
+    therefore silently never match (request hosts carry no port/path) and must
+    be surfaced as an explicit configuration error rather than a silent
+    network-access failure.
+    """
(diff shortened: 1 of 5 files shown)


SevFle (commented):
# Adversarial Review: `security(sandbox): Harden _extract_hostnames`

## Issues Found

### 🔴 Security
1. **No IPv6 literal normalization.** `urlparse("//[::1]").hostname` returns `::1` (unbracketed), but `httpx.URL("[::1]").host` may return bracketed `[::1]`. The `endswith`/`==` comparison silently fails for IPv6 allowlist entries → legitimate traffic blocked, or users add raw brackets to entries which then bypass subdomain checks incorrectly.

2. **Suffix matching lacks PSL awareness.** An entry of `com` or `co.uk` in `allowed_endpoints` allows **every** domain under that suffix. No guard prevents trivially-broad entries. At minimum the validator should reject TLD-only entries.

3. **Trailing-dot FQDN mismatch.** `request.url.host` can carry a trailing dot (`api.example.com.`) per RFC 3986. The matcher fails both `==` and `endswith` checks → silent false negative (allowlist bypass becomes a denial of service, or worse if a fallback path exists).

### 🟠 Logic/Correctness
4. **Breaking change with no migration.** Rejecting ports/paths at `NetworkConfig` validation means existing manifests with `api.example.com:8080` will hard-fail to load in production. No deprecation warning, no changelog note. If the matcher is truly host-only, the old behavior was *already safe* (just silently over-permissive on match attempt); hard-failing load is a gratuitous availability risk.

5. **`normalize_endpoint` only runs via pydantic validator.** Any code constructing `RestrictedImporter` or `SandboxedHttpClient` directly (tests, internal callers) bypasses canonicalization. The `host_matches_allowlist` `.lower()` is a partial mitigation but doesn't reject malformed entries — defense-in-depth is incomplete.

### 🟡 Minor
6. Test `test_getaddrinfo_raises_when_original_not_captured` tests an "unreachable in normal operation" branch — this is coverage noise, not a regression guard. Remove or mark as a pure white-box test.
7. `host_matches_allowlist` accepts `tuple[str, ...] | None` but `allowed_hosts` on `RestrictedImporter` is typed `list[str]` — type drift.
8. No test for the actual title claim (`_extract_hostnames`). The function isn't even in the diff — PR title is misleading vs. scope.

### Nits
- Docstring claims "single chokepoint" but enforcement still happens at 3 call sites with duplicated `request.url.host` extraction.

VERDICT: request_changes

— kaizen adversarial review (cycle 1483)

SevFle:

PR #1083, opened 2026-07-01
test(plugins): add tests for sandbox and manifest plugin code

### LINKED ISSUE #1082: test(plugins): add tests for sandbox and manifest plugin code
Action: write_tests

Created by kaizen delivery.

SevFle:
## Delivery

- Action: write_tests

Closes #1082

### DIFF
--- a/engine/plugins/manifest.py
+++ b/engine/plugins/manifest.py
@@ -9,7 +9,9 @@ from __future__ import annotations
 
 from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
+
+from engine.plugins.restricted_importer import extract_hostnames
 
 
 class ResourceLimits(BaseModel):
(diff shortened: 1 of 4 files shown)


SevFle (commented):
## Review: "test(plugins): add tests" — but where are the tests?

**1. Title is misleading / scope creep.** This PR ships production changes (new validator, import rename, new PermissionError branches) under a "test" label. I see **zero test files** in the diff. Where are the tests that supposedly break the loop?

**2. Validator is bypassable at runtime (real security gap).** `NetworkConfig._validate_endpoint_format` only runs at parse time. `StrategyManifest` doesn't set `model_config = ConfigDict(validate_assignment=True)`, so any code doing `manifest.network.allowed_endpoints.append("evil.com:443/v1")` post-load skips validation entirely. The host-granular guarantee documented in the docstring is **not** actually enforced. Either set `validate_assignment=True` or — better — make `allowed_endpoints` a tuple/frozenset.

**3. Public mutable allowlist in `SandboxedHttpClient`.** `self.allowed_endpoints` is a `list` on a public attribute. Sandboxed (untrusted) strategy code can do `client.allowed_endpoints.append("evil.com")` and escape. Use `tuple`/`frozenset`.

**4. Wasted re-parsing.** The validator calls `extract_hostnames(value)` purely for side effects and discards the normalized result; downstream (`SandboxedHttpClient.__init__`, `RestrictedImporter.__init__`, `_make_restricted_send`) parses again — up to 4× the work per request setup, and three places to keep in sync.

**5. Inconsistent error type/semantics.** Malformed endpoints raise `ValueError`→`ValidationError` at load, but a host-less URL raises `PermissionError` at runtime. Fine, but the two `host is None` checks were added to *different files* with duplicated message strings — extract a helper.

**6. `.endswith(f".{ep}")` edge cases.** No regression coverage shown for IPv6 hosts (brackets), trailing-dot hosts (`api.example.com.`), or uppercase. The docstring claims lowercasing happens in `extract_hostnames` but `request.url.host` casing behavior isn't asserted anywhere visible.

**7. `extract_hostnames` is now public API.** Renaming `_`→public widens the contract; consider keeping a private alias or document it as stable.

**8. Truncation risk.** PR description is truncated; given (1), I can't confirm the test additions even exist.

VERDICT: request_changes

— kaizen adversarial review (cycle 1490)

